PR #51, opened 2010-11-02
Escape closing tags in strings

mernen:
About the proposed patch: replacing the precise instances of "<&#47;script>" via a post-generation scan (like a `foo.gsub!("</script>", '<\/script>')`) would pretty much defeat the speed purposes of this gem (on the ext/jruby side), and replacing these instances mid-generation (through a look-ahead, for instance) would complicate the code too much. On the other hand, replacing every single "/" character with a "\/" sequence could make certain strings needlessly less readable (for instance, URLs starting with "http:&#47;/"). So, I think the compromise of replacing all "</" sequences with "<\/" offers the best combination of code simplicity and JSON readability. This patch implements this approach on all 3 versions (pure, ext, jruby) and adds a new test.


marcandre:
This should be closed in favor of #405

marcandre:
@hsbt could you please close this?

PR #98, opened 2011-09-27
Changes to work with Maglev implementation of Ruby C extensions.

timfel:
These changes will ensure that the official json gem is working on MagLev.

Major two problems fixed in these commits:
- Use RTEST() in conditionals, because Qfalse is not 0 on Maglev
- Use a macro to attach Ruby objects to the C json struct, that will, if compiled on Maglev, also set an ivar on that object, to keep the GC from collecting it. This is because in order to have an efficient GC Maglev cannot implement the rb_gc_mark\* functions or support the mark function for Data_Wrap_Struct.

If you're interested, see https://gist.github.com/1245462 if you want to know more about the changes needed for Maglev and the reasons for them


marcandre:
Sorry this wasn't addressed at the time.

Maglev targets Ruby 1.8.7 which we no longer support.

This can be closed (although we still don't call RTEST when we should...)

PR #166, opened 2013-03-22
Fix CFLAGS options in ext/json/ext/{generator,parser}/extconf.rb

vzvu3k6k:
`ext/json/ext/{generator,parser}/extconf.rb` have some problems.
1. [generator/extconf.rb](https://github.com/flori/json/blob/18261880e33b9bd70152d860ea88c784ebef07e4/ext/json/ext/generator/extconf.rb#L8-10) enables `-O0  -ggdb` when $DEBUG is false.
2. When $CFLAGS contains <code> -Ofast</code>, extconf.rb replaces it with `-O3fast`. It is invalid and causes compile error.

This patch fixes these two problems.

### DIFF
--- a/ext/json/ext/generator/extconf.rb
+++ b/ext/json/ext/generator/extconf.rb
@@ -1,13 +1,9 @@
 require 'mkmf'
 
-unless $CFLAGS.gsub!(/ -O[\dsz]?/, ' -O3')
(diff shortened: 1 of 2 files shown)


flori:
Fixed by deletion, actually.

PR #174, opened 2013-05-09
Keep Makefile out of git

mcandre:
I added `Makefile` to `.gitignore` to safely keep the generated makefile out of git version control.


marcandre:
Thanks for the PR.

Already ignored, this PR can be closed.

Issue #179, opened 2013-08-18
Workarounds to create ruby objects from unvalidated json data


movitto:
I recently updated from json 1.7.5 and ran into an issue caused by relying on the auto-conversion of json-data to ruby classes, as disabled by default in commit #d0a62f3c. I use json in the [rjr](https://github.com/movitto/rjr) gem where converting json data sent by end users is needed.

Inspecting the [CVE](https://groups.google.com/forum/#!msg/rubyonrails-security/4_YvCpLzL58/KBTICG7b4NgJ) and code, it seems the main culprit is the const lookup in lib/json/common.rb#deep_const_get. (there is also a note about GenericObject, but I'm not using that particular feature and json conversion is now disabled by default in that class anyways).

One partial solution would be to convert constants in the objects being searched to strings to compare to the user input, as opposed to converting the user input to a symbol as automatically done by the const_defined and const_get methods. Unfortunately we can't use const_missing in this fashion as there is no way to replicate that functionality without the symbol conversion. A small gist highlighting this can be seen [here](https://gist.github.com/movitto/6261634)

Another solution would be to employ a two stage parsing process, where the json data is securly converted to a hash, before the lookup occurs using a string registry. This in itself can be created ahead of time in a manner similar to json_creatable?, by invoking the functionality at the class level of classes that have been deemed secure. After the data is validated as being acceptable to convert, the actual conversion into ruby objects can be done (or an error thrown otherwise). Another small gist highlighting this can be seen [here](https://gist.github.com/movitto/6261662)

Just putting this out there to hopefully help others and open it up for comments as I decide what to do for the rjr gem. Neither solution is ideal but at least this would allow me to parse json data from the user while still addressing the security concerns. Also if there is any interest to get this in the json gem itself, I'd be happy to send a pull request.

Thanks,
  -Mo

byroot:
My plan short term is to issue a deprecation warning when the `create_addition` feature is used with `JSON.load`, and introduce some form of `JSON.unsafe_load`, which by the next major should be the only way to use that functionality.

it has been the source of many CVE over the years, we should get rid of that dangerous default.

So closing this.

PR #182, opened 2013-09-13
Add an optional create additions callback

movitto:
Follow up to workarounds describe in issue #179.

If not set create_additions still implements current behaviour.

If set to a proc this will be invoked in lieu of deep const get to evaluate the json class string and return a ruby class. It is up to the developer to specify the implementation details of this mapping.

Putting this out there for comments and hopeful merger into the codebase.

### DIFF
--- a/tests/test_json_addition.rb
+++ b/tests/test_json_addition.rb
@@ -101,6 +101,15 @@ class TestJSONAddition < Test::Unit::TestCase
     )
   end
 
+  def test_extended_json_custom_matcher
+    a = A.new(666)
+    assert A.json_creatable?
+    json = generate(a)
+    a_again = JSON.parse(json, :create_additions => proc { |e| A2 })
+    assert_kind_of A2, a_again
+    assert_equal a, a_again
+  end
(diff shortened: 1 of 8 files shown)


tests/test_json_addition.rb
```diff
@@ -101,6 +101,15 @@ def test_extended_json_disabled
     )
   end
 
+  def test_extended_json_custom_matcher
+    a = A.new(666)
+    assert A.json_creatable?
+    json = generate(a)
+    a_again = JSON.parse(json, :create_additions => proc { |e| A2 })
```
marcandre, line 108: A test is missing on what is `e`

byroot:
This PR diverged a bit too much to be realistically rebased. 

Also if anything I'm trying to discourage usage of `create_additions` as it has been the source of many vulnerabilities in the past.

That said, providing a proc is probably way cleaner than what is being done right now, and it might even help for performance, so it's not impossible I implement this myself in the near future. 

In the meantime I'm closing as non-actionable.

PR #192, opened 2013-12-29
JSON marshalling support for Set and SortedSet

jkanywhere:
Support JSON serialization of Set and SortedSet from Ruby standard library, in the style of your existing support for BigDecimal, Time, etc.
Allows the elements of a Set to be reconstructed.


AlexWayfer:
It looks added: https://github.com/flori/json/commit/06f93995137e0a61dd829fef3306fde262f88628

PR #193, opened 2013-12-29
Generation for additional types respect options

jkanywhere:
JSON serialization of `BigDecimal`, `Complex`, `Rational`, and `Regexp` now pass along
the options to handle, for example, pretty printing with `jj`.

From your documentation "don’t forget the *a for all the arguments" I think this is considered best practice, and many of the other classes you support, _e.g._ `Data`, `Exception`, already do this.


marcandre:
Already merged, can be closed.

PR #203, opened 2014-04-09
Perf. improvements to Hash#to_json in pure implementation generator.

vipulnsward:
Perf. improvements to Hash#to_json in pure implementation generator:
- Use `.each` instead of `for` for iterating over elements.
- Initialize `delim`, `result` vars with initial states instead of appending `state.object_nl` later.
- Use faster string interpolation where possible on `Hash#json_transform` instead of appending strings repeatedly.

Benchmark:

``` ruby
    hash ={:a => {:a => {:a => {:a => {:a => 1}}}}}
    require 'benchmark/ips'
    Benchmark.ips do |x|
      x.report("to_json ") {  hash.to_json}
    end
```

``` yaml
Before:
            to_json       4990 i/100ms
-------------------------------------------------
            to_json     52970.8 (±5.9%) i/s -     264470 in   5.010558s

After:
            to_json       5253 i/100ms
-------------------------------------------------
            to_json     55998.6 (±5.9%) i/s -     283662 in   5.083350s
```

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -248,7 +248,7 @@ module JSON
           else
             raise TypeError, "can't convert #{opts.class} into Hash"
           end
-          for key, value in opts
+          opts.each do |key, value|
             instance_variable_set "@#{key}", value
           end
           @indent                = opts[:indent] if opts.key?(:indent)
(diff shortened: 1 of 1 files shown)


marcandre:
This PR has conflicts.
There are some gains here, but they are pretty modest and I think the actually best way to improve the performance would be to build and array and `join` it at the very end (although I have not tried nor benchmarked this).
I would close this PR.

hsbt:
@vipulnsward Sorry, too late response. I update with the current HEAD and merged!

Issue #214, opened 2014-08-26
Generates incorrect JSON string for U+2028/U+2029


matthewbjones:
When the Ruby string contains a Unicode Line Separator character (http://www.fileformat.info/info/unicode/char/2028/index.htm), this library incorrectly generates the final JSON, which causes SyntaxError: Unexpected token ILLEGAL (in Chrome) and Unexpected EOF in Safari, etc.

``` ruby
"Home by Edward Sharpe & the Magnetic Zeros\u2028".to_json
# "\"Home by Edward Sharpe \\u0026 the Magnetic Zeros\u2028\""
```

should be

``` ruby
# "\"Home by Edward Sharpe \\u0026 the Magnetic Zeros\\u2028\""
```

As far as I can tell, the JSON spec and browsers treat U+2028/2029 as valid line separator characters, which are not allowed to occur within strings.

kyrias:
They aren’t allowed to occur in JavaScript strings, but they are perfectly valid in JSON strings.

marcandre:
IIUC, the JSON generated is correct, but the resulting is not Javascript code, contrary to what many people would expect. It is indeed preferable to escape those characters to produce equivalent JSON that is also valid Javascript code.

See also https://stackoverflow.com/questions/2965293/javascript-parse-error-on-u2028-unicode-character
Same issue in golang took 10 days to resolve... https://github.com/golang/go/issues/5836

byroot:
This is correct.

In any case, newer versions of the gem have a `script_safe` option for that use case.

PR #215, opened 2014-08-26
Ruby and C implementations now escape U+2028/2029.

matthewbjones:
This is a fix for Issue #214.

I modified the library to detect \u2028 and \u2029 within Ruby strings and escape those into their \u2028 and \u2029 counterparts. The test_json_unicode.rb was updated for these two conditions.


marcandre:
This PR looks good but need to be rebased and has conflicts.

Issue #206, opened 2014-05-26
Adapt json gem to conform to RFC 7159


duerst:
The JSON spec has been updated from RFC 4627 to RFC 7159 (https://tools.ietf.org/html/rfc7159). Among other things, this spec now allows for any JSON values at the top level. This also agrees with practice in Web browsers and with the ECMA spec. The Ruby JSON gem doesn't deal with this yet (see http://www.ietf.org/mail-archive/web/json/current/msg00670.html) and should be updated. There should most probably be an option to select either the old (RFC 4627) or the new (RFC 7159) behavior.

[There may also be other changes, but this is clearly the biggest one. Also, there's a profile of JSON called I-JSON in the works (see http://tools.ietf.org/html/draft-ietf-json-i-json). This currently keeps the restriction from RFC 4627, and adds some other sanity checks, but is not fully baked yet.]

Felk:
Any news on this? It's been almost 3 years since the new RFC and the library still fails to parse anything at root level that isn't a list or object

O1O1O1O:
@Felk the issue seems to be that quirks mode is default for JSON `load()` and `dump()` but not `parse()`.  If your input is from a trusted source you can always use `JSON.load("42")` or `JSON.load("true")`.  Better still just enabled quirks mode `JSON.parse("42", quirks_mode: true)`

Felk:
But... it's not a quirk, it's part of the standard :(

But that will do the job I guess, thanks

cabo:
No, quirks_mode is not doing the job; last time I looked it brought a lot of gunk with it that you don't want to have in a production system.  (I.e., it's not just a naming issue.)

I think we can all agree that the fact that the supposedly stable JSON spec changed after 10 years is unfortunate.  But it happened.  Now we have to find ways to deal with that.

Ruby needs a JSON parser.  Does this piece of software want to be it?  That would be good to know.

cben:
Since 7d2ad6d6556da03300a5aeadeeacaec563435773, quirks_mode param was removed, is now ignored.
But the current behavior (ruby 2.4.1) still seems to have 2 modes for other things:
```
[9] pry(main)> Float::NAN.to_json
JSON::GeneratorError: 865: NaN not allowed in JSON
from (pry):9:in `to_json'
[10] pry(main)> JSON.dump(Float::NAN)
=> "NaN"
[12] pry(main)> JSON.load("NaN")
=> NaN
[13] pry(main)> JSON.parse("NaN")
[...]

PR #220, opened 2014-11-03
RFC7159 mode

iainbeeston:
This adds a new "mode" to JSON which parse/generates JSON according to RFC7159 (the most up-to-date JSON spec), a fix for #206. The main difference between RFC4627 (the default in the JSON gem) and RFC7159 is that the latter RFC accepts JSON values as a valid JSON text.

You can use the new mode by passing `standards_mode: true` to `parse` (in a similar way to how `quirks_mode` is activated).


mpalmer:
It'd be nice to get this landed; RFC7159 isn't exactly brand new any more, and there's plenty of JSON out there that's using the new standard.

cabo:
Yes, please.  Before I have to convert my last tools to oj.

PR #236, opened 2015-02-13
backported ruby/ruby trunk r49562

hsbt:
@nobu fixed infinite recursion issue on ruby trunk. I create simple patch.


hsbt:
they are merged at bc33e102 and 4dec168abb84a9175549e8d03a95a57df6a384e7

PR #238, opened 2015-03-13
Setup travis to use ruby 2.2.1

ismaelga:
I just got a Segmentation fault on version 1.8.2 with ruby 2.2.1

Probably the fixes in the FFI might got something broken here.


ismaelga:
Hm, it actual seems that it's passing on ruby 2.2.1

ismaelga:
Never mind. This was in a new installed rails v4.2 ant it seems that spring was the culprit here.
Running `gem pristine --all` fixed it

flori:
I think the 2.2 in the travis config would fetch the newest 2.2.x ruby.

PR #250, opened 2015-06-12
store depth state on the stack.

tenderlove:
This removes mutations on the parser struct object and helps to
decouple the instance of the parser from parsing itself.

Eventually, if we can make the `JSON_ParserStruct` just a read-only
configuration object, then it could open the door to parser instance
reuse. (e.g. speed up situations where the configuation never changes,
like `loop { JSON.parse(somejson) }`)


tenderlove:
Looks like my `ragel` may be generating things differently than whatever one was used last.  I don't know if it's OK or not. :fearful:

PR #283, opened 2016-04-25
Remove comments from core classes in json/add

stomar:
_(Previously submitted to MRI's bug tracker, as https://bugs.ruby-lang.org/issues/12255)._

This patch removes some comments that otherwise would appear in MRI's docs of the affected core classes.

E.g. for `Exception`

```
$ ri Exception
Descendants of class Exception are used to [...]

[snipped class documentation]

Exception serialization/deserialization        <- does not belong here!
-------------------------------------------
Class methods:

[snip]
```

Or see for example http://docs.ruby-lang.org/en/trunk/Exception.html, where the same comment appears just above the "Public Class Methods" section.

### DIFF
--- a/lib/json/add/date.rb
+++ b/lib/json/add/date.rb
@@ -3,7 +3,6 @@ unless defined?(::JSON::JSON_LOADED) and ::JSON::JSON_LOADED
 end
 require 'date'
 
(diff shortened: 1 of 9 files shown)


stomar:
Thanks!

PR #313, opened 2017-01-12
Restore ruby 1.8 compatibility to json v1.8 branch

jeremyevans:
RB_TYPE_P is not available in ruby 1.8.7.  While json 2.0+ dropped
ruby 1.8 compatibility, the required_ruby_version in the gemspec
is not set (commented out), so the expectation is a tiny release
would not change which ruby versions are supported.


jeremyevans:
Looks like I should be using `rb_type` instead of `BUILTIN_TYPE`, so I pushed a new commit that does that.

marcandre:
Looks like `json` 1.x is no longer supported, sorry, so this should be closed.

PR #316, opened 2017-02-03
Output empty arrays as "[]" instead of "[\n\n]"

shreeve:


shreeve:
## Before this change:

```
$ ruby -rjson/pure -e 'puts JSON.pretty_generate([])'
[

]

$ ruby -rjson/pure -e 'puts JSON.pretty_generate([1])'
[
  1
]
```

## After this change:

```
$ ruby -rjson/pure -e 'puts JSON.pretty_generate([])'
[]

$ ruby -rjson/pure -e 'puts JSON.pretty_generate([1])'
[
  1
]
```

PR #327, opened 2017-04-12
Use assert_raise instead of assert_raises. 

hsbt:
It's not provide on ruby/ruby repository. :octocat:


hsbt:
I applied ruby/ruby at https://github.com/ruby/ruby/commit/d9df358ea8a627b9dab2ad2977f4d2448aad1e30

PR #333, opened 2017-06-01
Don't modify the encoding of an ASCII_8BIT string when it is parsed

cheister:
In the latest version of the gem a string that is not encoded as UTF-8 will be force_encoded to UTF-8 when it is parsed.

```
2.3.1 :001 > require 'json'
 => true 
2.3.1 :002 > JSON::VERSION
 => "2.1.0" 
2.3.1 :003 > json = "{}".force_encoding('ASCII-8BIT')
 => "{}" 
2.3.1 :004 > json.encoding.name
 => "ASCII-8BIT" 
2.3.1 :005 > JSON.parse(json)
 => {} 
2.3.1 :006 > json.encoding.name
 => "UTF-8" 
```

This wasn't the case with json 1.8.6

```
2.3.1 :001 > require 'json'
 => true 
2.3.1 :002 > JSON::VERSION
 => "1.8.6" 
2.3.1 :003 > json = "{}".force_encoding('ASCII-8BIT')
 => "{}" 
2.3.1 :004 > json.encoding.name
 => "ASCII-8BIT" 
2.3.1 :005 > JSON.parse(json)
 => {} 
2.3.1 :006 > json.encoding.name
 => "ASCII-8BIT" 
```

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -1685,9 +1685,7 @@ static VALUE convert_encoding(VALUE source)
 #ifdef HAVE_RUBY_ENCODING_H
   rb_encoding *enc = rb_enc_get(source);
   if (enc == rb_ascii8bit_encoding()) {
-    if (OBJ_FROZEN(source)) {
-      source = rb_str_dup(source);
-    }
+    source = rb_str_dup(source);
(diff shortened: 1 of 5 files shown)


marcandre:
Issue fixed in latest release, this can be closed.

PR #347, opened 2018-03-08
Enhance generic JSON and #generate docs

zverok:
What is here:

* adding docs for (questionable, yet present) "additions" feature;
* cleaning up (both formatting and formulations) docs of `JSON.generate`: current docs are poorly formatted and its documentation shifted more towards (internal) `State` object than just options;
* some other style/formatting cleanups.


zverok:
NB: I've rebased/fixed examples to the latest `master`

Issue #336, opened 2017-07-08
JRuby 9.2 with unified Integer raises ClassCastException


headius:
We have been working on our Ruby 2.4 support for a release in the next couple months. Rubhy 2.4 unified Fixnum and Bignum into Integer, which means some APIs that only expected to get Fixnum or Bignum before might need to adapt to support both.

It appears that the JRuby extension has not been updated for that change yet. I get the following error in MRI's test_json.rb:

```
[ 3/32] TestJSON#test_big_integers = 0.01 s
  1) Error:
TestJSON#test_big_integers:
Java::JavaLang::ClassCastException: org.jruby.RubyBignum cannot be cast to org.jruby.RubyFixnum
    json.ext.Generator$2.generate(Generator.java:224)
    json.ext.Generator$4.generate(Generator.java:292)
    json.ext.Generator$4.generate(Generator.java:249)
    json.ext.Generator$Handler.generateNew(Generator.java:171)
    json.ext.Generator.generateJson(Generator.java:57)
    json.ext.GeneratorState.generate(GeneratorState.java:209)
    json.ext.GeneratorState$INVOKER$i$1$0$generate.call(GeneratorState$INVOKER$i$1$0$generate.gen)
```

Any place that assumes it will only receive Fixnums should probably be audited. I can help, but I assume this work was already done for the C ext, and the changes should be very similar for the Java ext.

headius:
See jruby/jruby#4293 for ongoing Ruby 2.4 work. The branch is `ruby-2.4`.

skarmakar:
Seems like related, I tried this branch for this particular issue and getting the following error
```
jruby-9.2.5.0 :010 > {"a" => 1, "b" => 12345678901234567890}.to_json
Traceback (most recent call last):
       16: from org.jruby.ir.interpreter.InterpreterEngine.processCall(InterpreterEngine.java:346)
       15: from org.jruby.runtime.callsite.CachingCallSite.call(CachingCallSite.java:141)
       14: from org.jruby.runtime.callsite.CachingCallSite.cacheAndCall(CachingCallSite.java:325)
       13: from org.jruby.internal.runtime.methods.DynamicMethod.call(DynamicMethod.java:191)
       12: from org.jruby.internal.runtime.methods.JavaMethod$JavaMethodN.call(JavaMethod.java:793)
[...]

PR #355, opened 2018-06-02
Pass test_parse_big_integers on jruby

okazu-dm:
# Summary
I modified Generator.java and add INTEGER_HANDLER for passing test_parse_big_integers on jruby 9.2.0.0.
I think release of jruby 9.2.0.0 caused these failure. (I found this post http://jruby.org/2018/05/24/jruby-9-2-0-0) 

Maybe this problem is mentioned in #336 

# note of implementation
The test failed because of `Java::JavaLang::ClassCastException: org.jruby.RubyBignum cannot be cast to org.jruby.RubyFixnum`.
I avoid this excpetion by adding `INTEGER_HANDLER` and use this handler instead of `FIXNUM_HANDLER` and `BIGNUM_HANDLER`.

### DIFF
--- a/java/src/json/ext/Generator.java
+++ b/java/src/json/ext/Generator.java
@@ -7,12 +7,11 @@ package json.ext;
 
 import org.jruby.Ruby;
 import org.jruby.RubyArray;
-import org.jruby.RubyBignum;
 import org.jruby.RubyBoolean;
 import org.jruby.RubyClass;
-import org.jruby.RubyFixnum;
 import org.jruby.RubyFloat;
 import org.jruby.RubyHash;
+import org.jruby.RubyInteger;
 import org.jruby.RubyNumeric;
 import org.jruby.RubyString;
 import org.jruby.runtime.ThreadContext;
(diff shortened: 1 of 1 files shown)


hsbt:
@headius Can you review this?

kares:
this will only work on JRuby 9.2 (Ruby 2.5) compat but will break under 9.1 (Ruby 2.3).
... so whether this is okay depends on whether the release would target ruby >= 2.5

kares:
setup an alternative fix, which works all the way back to JRuby 1.7 ... https://github.com/flori/json/pull/361

kares:
just to be clear, this piece did not raise any failures 
but Fixnum/Bignum objects did take the generic path with the `return GENERIC_HANDLER;` on JRuby < 9.2 ... which isn't ideal/intended

marcandre:
Commits should be squashed.

I can't set the reviewer to @headius (or maybe another JRuby user can review?)

kares:
Hey Marc, this PR is defunct due https://github.com/flori/json/pull/361 being merged ... there's special handling for the internal `RubyFixnum`/`RubyBignum` objects used by JRuby (these won't surface as a warning to use `Integer`).

okazu-dm:
@kares  Oh I found #361 fix this problem now. I close this PR.
Thanks for your kindness.

PR #360, opened 2018-10-31
Removed control characters from gemspec.

hsbt:


eregon:
Does upstream RubyGems support separating those paths by space though?
It would be good to verify it still works as intended (see https://github.com/rubygems/rubygems/issues/883).

PR #362, opened 2018-12-02
Fix for bigdecimal updates

mrkn:
`BigDecimal.new` is no longer available from bigdecimal-1.4.0.
This pull request removes the dependencies that method.


mrkn:
Note that the same change has been applied in ruby-trunk.

PR #377, opened 2019-03-19
Remove useless `stub` headers from .gemspec.

voxik:
The headers are generated by RubyGems in an installed .gemspec file but have no meaning in the source .gemspec file.


hsbt:
@voxik I removed control-chars at https://github.com/flori/json/pull/398 . Can you rebase this from the current master branch? Thanks.

hsbt:
@voxik ping

marcandre:
This should be closed at it is not the correct solution. I propose #415 instead.

hsbt:
make sense. Thanks.

Issue #386, opened 2019-10-22
Provide a `load_file` method like YAML's


keithrbennett:
One very common use case is to need to parse the contents of a JSON file.

As far as I know, this can only be accomplished in 2 steps, e.g. `JSON.parse(File.read('something.json'))`.

YAML provides a convenient `load_file` method that combines these two. Obviously, it is trivially simple to implement.

Could we get a similar method (preferably with the same name) in this JSON library?

hsbt:
https://github.com/flori/json/pull/387 was merged now.

PR #387, opened 2019-10-22
Add `load_file` and `load_file!` methods, with tests. Fixes issue #386.

keithrbennett:
Adds `load_file` methods analogous to YAML's.

### DIFF
--- a/lib/json/common.rb
+++ b/lib/json/common.rb
@@ -282,6 +282,16 @@ module JSON
     Parser.new(source, **(opts||{})).parse
   end
 
+  # Parses the content of a file (see parse method documentation for more information).
+  def load_file(filespec, opts = {})
+    parse(File.read(filespec), opts)
+  end
+
+  # Parses the content of a file (see parse! method documentation for more information).
+  def load_file!(filespec, opts = {})
+    parse!(File.read(filespec), opts)
+  end
+
   # :call-seq:
   #   JSON.generate(obj, opts = nil) -> new_string
   #
--- a/tests/json_common_interface_test.rb
+++ b/tests/json_common_interface_test.rb
@@ -123,4 +123,60 @@ class JSONCommonInterfaceTest < Test::Unit::TestCase
     assert_equal @json, JSON(@hash)
     assert_equal @hash, JSON(@json)
   end
+
+  def test_load_file
+    test_load_shared(:load_file)
+  end
+
+  def test_load_file!
+    test_load_shared(:load_file!)
+  end
+
+  def test_load_file_with_option
+    test_load_file_with_option_shared(:load_file)
+  end
+
+  def test_load_file_with_option!
+    test_load_file_with_option_shared(:load_file!)
+  end
+
+  private
+
+  def test_load_shared(method_name)
+    temp_file_containing(@json) do |filespec|
+      assert_equal JSON.public_send(method_name, filespec), @hash
+    end
+  end
+
+  def test_load_file_with_option_shared(method_name)
+    temp_file_containing(@json) do |filespec|
+      parsed_object = JSON.public_send(method_name, filespec, symbolize_names: true)
+      key_classes = parsed_object.keys.map(&:class)
+      assert_true key_classes.include?(Symbol) && (! key_classes.include?(String))
+    end
+  end
+
+  # Copied and slightly modified from https://github.com/keithrbennett/trick_bag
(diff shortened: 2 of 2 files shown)


lib/json/common.rb
```diff
@@ -179,6 +179,20 @@ def parse!(source, opts = {})
     Parser.new(source, opts).parse
   end
 
+
+  # Parses the content of a file
+  # See parse method documentation for more information.
+  def load_file(filespec, opts = {})
+    parse(File.read(filespec))
```
rafaltrojanowski: @keithrbennett I'm just wondering, shouldn't we pass options for 'File.read'? It might be useful for supporting different encodings, etc. It's not related but I have a case where I struggle with BOM character in the beginning of the CSV file and by passing encoding I can simply remove those additional bytes. Maybe that would work similarly in JSON.

keithrbennett: First I want to thank you for pointing out to me that I was failing to pass the passed `opts` along to the called methods. I've fixed that and pushed the fix.

Regarding your point about `File.read` options, do I understand you correctly that you would want to specify options to `File.read`, not `JSON.parse`? If so, I'm not sure how we could do that, since we are already using the Ruby hash last parameter feature for the `JSON.parse` options.

Any strategy to handle options to `File.read` would need to be balanced with the cost of the complexity added to support it. Can you suggest a way to accommodate this need in the least complex way?

rafaltrojanowski:
Thanks @keithrbennett!

lib/json/common.rb
rafaltrojanowski: Yes, but I'm happy with current solution. I don't have idea how to solve it in a elegant way. Maybe maintainer can help. And sorry for the oftop, anyway it's better than before :)

tests/json_common_interface_test.rb
```diff
@@ -123,4 +123,43 @@ def test_JSON
     assert_equal @json, JSON(@hash)
     assert_equal @hash, JSON(@json)
   end
+
+
+  def test_load_file
```
marcandre, line 128: Please add a test with options...

lib/json/common.rb
marcandre: Best not mix options. If a user need options, they can do the `File.read`. FWIW, `YAML.load_file` has no file option either.

tests/json_common_interface_test.rb
keithrbennett: Sorry about that. I've added this (and an equivalent function for `load_file!`), would it suffice?:

```
  def test_load_file_with_option
    temp_file_containing(@json) do |filespec|
      key_classes = load_file(filespec, symbolize_names: true).keys.map(&:class)
      assert_true (key_classes.include?(Symbol) && (! key_classes.include?(String)))
    end
  end
```

Also, in order to get the tests to pass I had to `rebase master` my feature branch. Is it ok to `git push -f` this branch?

marcandre: > Is it ok to git push -f this branch?

Yes, that's the best thing to do 😄

marcandre:
Very good. Only thing remaining would be to squash your commits (although it can be done by maintainers with 'squash and merge')

tests/json_common_interface_test.rb
```diff
@@ -123,4 +123,61 @@ def test_JSON
...
+
+  def test_load_file_with_option!
+    test_load_file_with_option_shared(:load_file!)
+  end
+
+  private
+
+  def test_load_shared(method_name)
+    temp_file_containing(@json) do |filespec|
+      assert_equal method(method_name).call(filespec), @hash
```
marcandre, line 148: `method(method_name).call(filespec)` => `public_send(method_name, filespec)`. I think it's more idiomatic.
Same for `test_load_file_with_option_shared`

keithrbennett: I've made this change and squashed the commits. Thanks for your guidance, I agree that these are both improvements.

marcandre: Thanks for the impressively quick turnaround and good PR!

hsbt:
@keithrbennett Thanks. Could you rebase from this from master branch? 

I will merge this targeted json-2.4.0 after releasing json-2.3.1.

keithrbennett:
Done, thanks.

- Keith

hsbt:
Thanks!

PR #401, opened 2020-01-12
Add option :allow_trailing_comma to JSON#parse

jspanjers:
I think it would be really nice to support JSON arrays with trailing  comma's!

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -9,14 +9,14 @@
 static void
 enc_raise(rb_encoding *enc, VALUE exc, const char *fmt, ...)
 {
-	va_list args;
-	VALUE mesg;
+    va_list args;
+    VALUE mesg;
 
-	va_start(args, fmt);
-	mesg = rb_enc_vsprintf(enc, fmt, args);
-	va_end(args);
+    va_start(args, fmt);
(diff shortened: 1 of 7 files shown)


marcandre:
Nice PR 🎉

I don't know if this will be accepted (since this goes beyond the stated goal of sticking to the RFC).

If not, you should consider making a [json5](https://json5.org/) parser. The [one I know about](https://github.com/bartoszkopinski/json5) has 500k downloads despite saying "Experimental JSON5 Ruby parser - DO NOT USE". The specs are already written too :-)

marcandre:
Note: we already deviate from the RFC, for example with NaN, so I think this PR should be accepted.

marcandre:
@hsbt: can we allow this feature?

jspanjers:
Hi there,

I've rebased my feature branch to the current master to resolve some merge conflicts.
I have also added the implementation for JRuby.

Please review this pull request again and/or approve running the workflow to run the tests.

byroot:
I think we should do this. As pointed out, we already have an option to support `NaN` and `Infinity`, and we even support comments with no option to turn it off.

Trailing commas would be the only thing missing to be compatible with what's often referred to as "jsonc" (some `jsonc` specs also allow unquoted keys, but that's going a bit far).

I'll try to rebase this PR.

PR #402, opened 2020-01-16
Test with TruffleRuby in CI

eregon:
The latest release of JSON (2.3.0) unfortunately broke TruffleRuby 19.3.0, as it used `rb_gc_register_mark_object` which was not yet supported by TruffleRuby: https://github.com/oracle/truffleruby/pull/1856
The TruffleRuby 19.3.1 release fixes it.

It would be useful to test TruffleRuby in CI to notice when it breaks, and if possible avoid breaking it.

I checked and it [fails as expected](https://travis-ci.org/flori/json/jobs/637792223) when using a not-yet-implemented C API function.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,6 @@ matrix:
     - rvm: ruby-head
     - rvm: jruby
     - rvm: jruby-9.2.7.0
+    - rvm: truffleruby
 script: "bundle exec rake"
 sudo: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,7 @@ rvm:
   - ruby-head
   - jruby
   - jruby-9.2.7.0
+  - truffleruby
 matrix:
   allow_failures:
     - rvm: ruby-head


eregon:
@hsbt Could you review and merge this PR?

eregon:
@nurse @hsbt @flori Ping, could you merge this?

hsbt:
I'm negative to merge this. Because we didn't support TruffleRuby officially yet.

At first, we should decide before merging this.

eregon:
> At first, we should decide before merging this.

Who can decide that?

JSON is part of the standard library and widely used, it is critical that it works on all major Ruby implementations, that is MRI, JRuby and TruffleRuby.

Supporting TruffleRuby currently requires 0 changes in the gem (unlike JRuby which needs a completely different extension). I think it doesn't make sense to support JRuby but not TruffleRuby in that context.

I want to clarify that merging this PR doesn't mean the new TravisCI TruffleRuby job must prevent merging changes.
But if it fails, it should be a conscious decision (i.e., not accidental) and of course it will be great if someone from TruffleRuby knows about it. I can take care of monitoring the CI status here for that.

eregon:
@hsbt How can we progress on this PR? What should we wait for?

I'm not asking for official support of TruffleRuby in JSON, just adding it in CI.

FWIW, I'm happy to fix any issue coming with TruffleRuby in JSON.
We can just move truffleruby to `allow_failures` if it fails (obviously would be great if I know about that, but I can also add some automated monitoring on my side).

Given that, can we just merge this PR?

nurse:
In general who can decide this is flori.

eregon:
@flori Could you review and merge this PR?

It would be great to have other maintainers for such a critical part of the Ruby standard library.
Can we consider @hsbt @nurse and @headius as maintainers alongside you, which means they should be able to make all decisions needed to maintain this project? (including merging PRs, adding support for new implementations, publish releases, push the gem, etc).
From https://bugs.ruby-lang.org/issues/15605 and many reports before, it's clearly not enough to have a single not-so-active maintainer.
I don't think it's sensible to require the project creator for such a small PR. Therefore I'd like to address the general situation here.

Maybe transferring the project to https://github.com/ruby/json would help to clarify this is a Ruby community-maintained gem. Most other gems in the standard library are under https://github.com/ruby, so I think it would be consistent.

@flori What do you think of that? Of course, we can move this discussion to an issue or somewhere else if you prefer.

MSP-Greg:
See PR https://github.com/flori/json/pull/406 which add Actions CI for all platforms.  TruffleRuby is  passing on both Ubuntu & macOS, along with all MRI Rubies 2.2 thru head, on all three platforms...

eregon:
@flori This PR has been open for 2 months, could you please take a look at it?

headius:
I would merge if I had push rights, but I do not.

hsbt:
@eregon Can you also add the truffleruby to `allow_failures`? After that, I'm going to merge this.

> We can just move truffleruby to `allow_failures` if it fails (obviously would be great if I know about that, but I can also add some automated monitoring on my side).

It's your convenience, not maintainer's one. I was always tired to maintain CI matrix for no responsible platform.

eregon:
@hsbt Done.

I think it's not much effort for the potential PR breaking TruffleRuby to add it in `allow_failures`, but I understand it might be annoying e.g., when synchronizing commits from ruby/ruby.
I'll find a way to monitor the result of TravisCI on this repo so I don't miss it if it breaks.

hsbt:
@eregon Thanks!

eregon:
Thanks for merging!

@flori Could you reply to https://github.com/flori/json/pull/402#issuecomment-586585888 once you see this?

PR #405, opened 2020-01-30
Add an option to escape forward slash character

casperisfine:
This is a rebase of https://github.com/flori/json/pull/235.

> As per the spec, the forward slash should be escaped in json strings. Not escaping the forward slash causes a security issue when a json object is interpolated inside of a script tag.

> For instance:

```ruby
irb(main):013:0> include ActionView::Helpers::TagHelper
=> Object
irb(main):014:0> include ActionView::Helpers::JavaScriptHelper
=> Object
irb(main):015:0> puts javascript_tag 'var json='+JSON.generate({"key"=>"</script><script>whatever()//"})
<script>
//<![CDATA[
var json={"key":"</script><script>whatever()//"}
//]]>
</script>
```
> In javascript `</script>` has precedence over everything else, so in this case the script tag is terminated and `whatever()` is executed even though it should be part of the string.

> The correct way to defend against this security issue is to escape the forward slash, so that <\/script> does not terminate the original script tag and the json string stays a string.

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -22,7 +22,7 @@ static ID i_to_s, i_to_json, i_new, i_indent, i_space, i_space_before,
           i_object_nl, i_array_nl, i_max_nesting, i_allow_nan, i_ascii_only,
           i_pack, i_unpack, i_create_id, i_extend, i_key_p,
           i_aref, i_send, i_respond_to_p, i_match, i_keys, i_depth,
-          i_buffer_initial_length, i_dup;
+          i_buffer_initial_length, i_dup, i_escape_slash;
 
 /*
  * Copyright 2001-2004 Unicode, Inc.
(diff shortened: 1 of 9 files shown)


casperisfine:
@hsbt any chance this could be considered?

marcandre:
If I'm not mistaken, the default for `escape_slash` is `false` in general, but it is `true` for `j`, `jj`, `dump_default_options` and thus `JSON.generate`? I don't know if that the best default.

casperisfine:
Thanks for the review. I'll update the PR soon.

casperisfine:
@marcandre all done!

casperisfine:
> If I'm not mistaken, the default for `escape_slash` is `false` in general, but it is `true` for `j`, `jj`, `dump_default_options` and thus `JSON.generate`? I don't know if that the best default.

I just pushed another commit to address this. These were leftovers from #235 which made `escape_slash` enabled by default, but that was deemed a compatibility breakage.

IMHO it would be preferable to enable it by default at some point, but I understand how it's better to be cautious. So let's keep it purely opt-in for now.

marcandre:
Documentation has been greatly improved; @casperisfine could you resolve the conflicts?

casperisfine:
Done.

marcandre:
Thanks @casperisfine :+1:

@hsbt thank you in advance for merging this

casperisfine:
Any news on this?

marcandre:
@hsbt can this be merged?

hsbt:
I have no opnion about this.

@nurse Can you review this?

nurse:
As flori said in https://github.com/flori/json/pull/235#issuecomment-96826256, he agreed this feature if it's configurable and disabled by default. Now this PR follows the comment and it's considered as acceptable. I merge this.

PR #406, opened 2020-03-18
Add GitHub Actions CI

MSP-Greg:
Two commits:

1. 'Rakefile - update for Windows mingw & mswin' - minor Windows specific changes to allow compiling on mingw & mswin

2. 'Add GitHub Actions - test Ubuntu, macOS, Windows' - two workflow files

    1. Pass - Test MRI Rubies, 2.2 thru head, on Ubuntu, macOS, & Windows.  Test TruffleRuby on Ubuntu & macOS.

    2. Fail - Test JRuby 9.2 and JRuby head on Ubuntu & macOS.  All jobs are 'allow-failure'.

### DIFF
--- a/.github/workflows/json.yml
+++ b/.github/workflows/json.yml
@@ -1,0 +1,66 @@
+name: Json
+
+on: [push, pull_request]
+
+jobs:
+  pass:
+    name: >-
+      pass ${{ matrix.os }} ${{ matrix.ruby }}
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu, macos, windows ]
+        ruby: [ head, 2.7, 2.6, 2.5, 2.4, 2.3, 2.2, truffleruby ]
+        include:
+          - { os: windows, ruby: mingw }
+          - { os: windows, ruby: mswin }
+        exclude:
+          - { os: windows, ruby: head }
+          - { os: windows, ruby: truffleruby }
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: MSP-Greg/setup-ruby-pkgs@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          apt-get: ragel
+          brew: ragel
+          mingw: _upgrade_ ragel
+
+      - name: Install dependencies
+        run:  bundle install
+
+      - name: Compile
+        run:  rake compile
+
+      - name: Test
+        run:  rake test
+
+  fail:
+    name: >-
+      fail ${{ matrix.os }} ${{ matrix.ruby }}
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu, macos ]
+        ruby: [ jruby-head, jruby ]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: MSP-Greg/setup-ruby-pkgs@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          apt-get: ragel
+          brew: ragel
+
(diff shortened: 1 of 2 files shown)


MSP-Greg:
The main reason I submitted this PR was to achieve parity with the MRI platforms used in the main ruby/ruby CI.  Having CI pass here and then cause CI failures in ruby/ruby is a PITA.  The Ruby master builds used in Actions CI are fully tested, unlike the builds on other CI platforms.

I've separated the non MRI CI into a separate workflow file and all fail, so they're set 'allow-failure'.  More can be added, like TruffleRuby...

MSP-Greg:
Rebased and updated to make use of changes to Actions matrix api

.github/workflows/json.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  pass:
+    name: >-
+      pass ${{ matrix.os }} ${{ matrix.ruby }}
+    runs-on: ${{ matrix.os }}-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ ubuntu, macos, windows ]
```
hsbt, line 13: Do not mix Linux/macOS and Windows platforms in the one workflow.

.github/workflows/json.yml
```diff
@@ -0,0 +1,66 @@
...
+        include:
+          - { os: windows, ruby: mingw }
+          - { os: windows, ruby: mswin }
+        exclude:
+          - { os: windows, ruby: head }
+          - { os: windows, ruby: truffleruby }
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Ruby
+        uses: MSP-Greg/setup-ruby-pkgs@v1
```
hsbt, line 24: Can you remove the your actions? The maintainers can't control it.

.github/workflows/json.yml
eregon: I am curious, why do you prefer separate workflows?
I usually prefer like this if there isn't much OS-specific logic, as it avoids duplication and makes it significantly easier to maintain (1 place to change vs 3).

.github/workflows/json.yml
eregon: It's needed to install `ragel` in a platform-agnostic way.
`MSP-Greg/setup-ruby-pkgs@v1` just uses `ruby/setup-ruby` for the setting up Ruby part.

MSP-Greg:
@hsbt

> Can you remove the your actions? The maintainers can't control it.

I'm happy to do one or both of the following:
1. Add additional maintainers
2. Move it to another account

.github/workflows/json.yml
hsbt: The maintainability is low about mixing *nix and Windows for my experience.

hsbt:
> I'm happy to do one or both of the following:

I have no motivation about that.

hsbt:
Unfortunately, Greg did reject my suggestion.

eregon:
@hsbt I don't understand why you closed this PR, could you elaborate?

What's the issue with `setup-ruby-pkgs`? As far as I can see here @MSP-Greg offered to add additional maintainers or move it to another account, but you didn't say if you wanted any of that or something else.

Also I don't see Greg rejecting any suggestion here.

.github/workflows/json.yml
eregon: That may have been true in the past at the beginning of GitHub Actions, but I don't think it's the case anymore, especially with improvements in ruby/setup-ruby. In fact many gems already use such an approach and in my experience, it's successful.

From a maintainability point of view it certainly looks better to me to have the steps repeated the minimum number of times (2 times here for allowed failures, instead of 3-4 if we have one job definition per OS).

hsbt:
He refused the maintainer's requests this and https://github.com/ruby/psych/pull/444. I never collaborate these attitude.

PR #418, opened 2020-06-25
Sync ruby

nobu:


hsbt:
Thanks!

voxik:
Since the .gemspec is autogenerated by Rake, does the change from 9c29ce3 survive next `rake gemspec` run? I don't think so (have not tested myself).

marcandre:
@voxik Currently `json_pure.gemspec` is generated and not `json.gemspec`, so this is fine.

PR #421, opened 2020-06-28
Add Time#to_json method to improve performance

Watson1978:
When it pass an object containing Time objct as an argument of `JSON.generate`,
the following processes are performed.

1. it will invoke `Time#to_json`
   at https://github.com/flori/json/blob/3d1c6a9246251e382c80dd8d85844c0bbf182da1/ext/json/ext/generator/generator.c#L1019
2. `mObject_to_json` function will be called by invoking `Time#to_json`.
3. it will invoke `Time#to_s`
   at https://github.com/flori/json/blob/3d1c6a9246251e382c80dd8d85844c0bbf182da1/ext/json/ext/generator/generator.c#L576

Seems that it reducde the performance by two Ruby method calls using `rb_funcall`.

This patch adds `generate_json_time` and makes a direct call to `Time#to_s` to reduce `rb_funcall` calling.
The performance is improved by about 20 %.

## Environment
- MacBook Pro (16-inch, 2019)
- macOS 10.15.5
- Intel Core i9 2.4 GHz

### Before
```
       JSON.generate    16.650k i/100ms
        Time.to_json    20.176k i/100ms
       JSON.generate    167.636k (± 0.9%) i/s -    849.150k in   5.065832s
        Time.to_json    200.217k (± 1.2%) i/s -      1.009M in   5.039224s
```

### After
```
       JSON.generate    20.297k i/100ms
        Time.to_json    20.329k i/100ms
       JSON.generate    205.745k (± 1.1%) i/s -      1.035M in   5.031798s
        Time.to_json    203.763k (± 0.7%) i/s -      1.037M in   5.088436s
```

### Test code
```ruby
require 'json'
require 'benchmark/ips'

time = Time.now

Benchmark.ips do |x|
  x.report('JSON.generate') { JSON.generate(time) }
  x.report('Time.to_json') { time.to_json }
end
```

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -13,7 +13,7 @@ static VALUE mJSON, mExt, mGenerator, cState, mGeneratorMethods, mObject,
 #else
              mFixnum, mBignum,
 #endif
-             mFloat, mString, mString_Extend,
+             mFloat, mTime, mString, mString_Extend,
              mTrueClass, mFalseClass, mNilClass, eGeneratorError,
              eNestingError,
              i_SAFE_STATE_PROTOTYPE;
@@ -464,6 +464,16 @@ static VALUE mBignum_to_json(int argc, VALUE *argv, VALUE self)
 static VALUE mFloat_to_json(int argc, VALUE *argv, VALUE self)
 {
     GENERATE_JSON(float);
+}
+
+/*
+ * call-seq: to_json(*)
+ *
+ * Returns a JSON string representation for this Time object.
+ */
+static VALUE mTime_to_json(int argc, VALUE *argv, VALUE self)
+{
+    GENERATE_JSON(time);
 }
 
 /*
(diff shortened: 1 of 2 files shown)


marcandre:
Cool PR. I find that gain surprising. I'm not against optimizing for `Time`, but it would be even nicer to optimize for all objects without a specialized `to_json`...

So one thing I notice in `generate_json` is the call to `rb_respond_to(obj, i_to_json)`. Since this will be true for the vast majority of cases, couldn't we instead simply call `to_json` on the `obj` and only in case we get a `NoMethodError` have the fallback that follows?

In Ruby:
```ruby
# generate_json
case
when obj.is_a?(Hash)
  # ...
when obj.respond_to?(:to_json)
  obj.to_json
else
  obj.to_s
end
# change to:
case
when obj.is_a?(Hash)
  # ...
else
  begin
    obj.to_json
  rescue NoMethodError
    raise if obj.repond_to?(:to_json) # must be an error within a custom `#to_json`
    obj.to_s
  end
end
```

I'm curious to know much of that 20% gain is due to this avoidable call to `respond_to?`?

marcandre:
And now that I'm thinking about it, it's a bit odd that there's not a `rb_call_if_respond_to` or similar, to avoid this pattern of looking up for a method call twice.

Watson1978:
```
  begin
    obj.to_json
  rescue NoMethodError
    raise if obj.repond_to?(:to_json) # must be an error within a custom `#to_json`
    obj.to_s
  end
```

In the case of calling #to_s, I'm guessing that performance will be worse than currently.
(I think it is big overhead to catch the exceptions.)

So, I think it would be better to verify that in another PR.

marcandre:
> In the case of calling #to_s,

No, in the case of `generate_json` which will not call `to_s` unless the object inherits only from BasicObject or has `undef to_json`.

> (I think it is big overhead to catch the exceptions.)

I'm sorry if I wasn't clear, but we wouldn't be catching exceptions (except in the very unusual cases of an object that inherits only from BasicObject or has `undef to_json`).

Or am I missing something?

PR #424, opened 2020-06-30
Update Changes for 2.3.0 [Fixes #397]

marcandre:
This is a summary of the changes to v2.3.0 which I hope is accurate.

Issue #409, opened 2020-04-16
create_additions defaults to true when using JSON.load


leanne73:
Currently in the process of updating from version 1.8.6 to 2.3.0, and the following code:
```
JSON.load("{\"hello\":\"world\"}", nil, symbolize_names: true)
```
now raises this error
```
ArgumentError: options :symbolize_names and :create_additions cannot be  used in conjunction
from /usr/local/bundle/gems/json-2.3.0/lib/json/common.rb:156:in `initialize'
```
From the docs, it looks like `create_additions` should be defaulting to `false` but now is not. I can work around the issue by appending `create_additions: false` to the method call, but it seems like the extra parameter shouldn't be necessary.

Note: `JSON.parse("{\"hello\":\"world\"}", symbolize_names: true)` still works as expected. It looks like the issue is only with JSON.load

msbit:
Had the same issue myself.

Looking over the history of this, it looks like `create_additions` _might_ have been incorrectly set as default as part of https://github.com/flori/json/commit/d0a62f3ced7560daba2ad546d83f0479a5ae2cf2?

marcandre:
The documentation for `load_default_options` [has been fixed] and states that `create_additions` is `true` by default. So it is expected that you have to set `create_additions` to `false`.

Let me remind you that the doc states clearly about `JSON.load`: "BEWARE: This method is meant to serialize data from trusted user input".

In general, you should be using `parse`, not `load`.

HTH

This issue can be closed.

msbit:
@marcandre looks like there may be another spot where the documentation needs to be updated:

https://github.com/flori/json/blob/98db309f5e5fd37f7c81b511170973fc6d99ed2a/lib/json/pure/parser.rb#L70

marcandre:
@msbit I believe the documentation is correct. If you create a `Parser` from scratch (pure Ruby or C), the default is `false`.

It's only if you call `JSON.load`, then it takes the `load_default_options` and [merges them with the options you give it](https://github.com/flori/json/blob/master/lib/json/common.rb#L559)

msbit:
@marcandre oh, okay, fair enough :)

leanne73:
Alright, closing as the documentation has been updated now.

PR #428, opened 2020-06-30
Make changes more precise [#424]

marcandre:
Sorry @hsbt, issue #409 made me realize it's best to make my wording more precise, as `JSON.load` still has default set to true.
Thanks in advance.


hsbt:
👍 

It's useful explanation for CVE. Thanks.

PR #430, opened 2020-06-30
Move options from #generate and #parse to common area

BurdetteLamar:
This move is so that all methods (load, dump, etc.) can refer to the options, not just generate and parse.


BurdetteLamar:
@hsbt, I don't want to do more before this is merged.  Can you take a look?  Thanks.

PR #435, opened 2020-07-06
unify json-java gemspec with the baseline

kares:
json gemspec requires Ruby 2.x while the -java platform gem does not.

doing so we could also drop some ugly code present due JRuby 1.7 (Ruby 1.9.3) compatibility.

HINT: did some (logical) re-ordering in the gemspec to "match" parts, 
think the gemspec do not need to be separate but keeping things as is, for now.


hsbt:
👍 It's good cleanup.

PR #436, opened 2020-07-07
[test] properly 'skip' test on JRuby

kares:
an early `return` still caused [ensure to execute](https://github.com/kares/json/blob/1e31f59231d7d31dc58a1e07f228897ddd4b4d7b/tests/json_generator_test.rb#L76),
setting `JSON::SAFE_STATE_PROTOTYPE` constant to `nil` for later tests!


marcandre:
👍

PR #439, opened 2020-07-28
RDoc for JSON.load with proc

BurdetteLamar:
There was no example of load with proc (and there's no test).


marcandre:
Thanks @hsbt 👍

PR #442, opened 2020-07-29
Nodoc for recurse_proc

BurdetteLamar:
Suggest :nodoc: for method recurse_proc.  It's called only (and optionally) by method load, and may not be useful to JSON users.

If we don't :nodoc: this method, it should at least have some tests, which I think are now lacking.

### DIFF
--- a/lib/json/common.rb
+++ b/lib/json/common.rb
@@ -561,7 +561,7 @@ module JSON
   end
 
   # Recursively calls passed _Proc_ if the parsed data structure is an _Array_ or _Hash_
-  def recurse_proc(result, &proc)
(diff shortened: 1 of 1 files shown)


BurdetteLamar:
@marcandre, does this seem right to you?

marcandre:
Mmm, that method is public but not clearly related to json.

Marking it as `:nodoc:` seems reasonable to me.

BurdetteLamar:
@hsbt, can you take a look?

BurdetteLamar:
Same for JSNONiconv and JSON.valid_utf8?, yes?  If so, please mod here.

PR #444, opened 2020-08-14
bundle the LICENSE file in the gem

jfeltesse-mdsol:
From https://guides.rubygems.org/specification-reference/#license=

> (...) The full text of the license should be inside of the gem (at the top level) when you build it.

Oddly enough, `json_pure.gemspec` does include the file but not the others.

Also, `json-java.gemspec` called `licenses=` twice.


hsbt:
Thanks.

PR #446, opened 2020-08-20
Partial compliance with doc/method_documentation.rdoc

BurdetteLamar:
Complies with https://github.com/ruby/ruby/blob/master/doc/method_documentation.rdoc#argument-description-if-necessary- by removing references to *-convertible thingies.


marcandre:
LGTM. @hsbt thanks for merging when you have a second.

PR #447, opened 2020-09-09
Implement a freeze: parser option

casperisfine:
If set to true all parsed objects will be immediately frozen, and strings will be deduplicated if the Ruby implementation allows it.

This option is useful if you plan to keep the parsed objects around, and want to reduce their memory footprint, and you know they contain duplicated string. It's common in says I18n data, or some database like configuration files (e.g. tax rates etc). 

A similar option was added to YAML/Pysch: https://github.com/ruby/psych/pull/414

MessagePack is also about to have it: https://github.com/msgpack/msgpack-ruby/pull/194 (at least the principle was accepted).

Right now this option incur a small overhead because it has to call `String#-@` through `rb_funcall`, but hopefully Ruby 3.0 should soon expose `rb_fstring_new(const char *ptr, long len) (likely under another name)` which should allow to reduce allocations by directly looking up existing strings without allocating a new one. 

NB: I presume that a java implementation is needed, I'm not too well versed into Java by I can try my hand at it. However I'd rather have your feedback on the MRI and "pure" implementations first.

### DIFF
--- a/ext/json/ext/parser/parser.rl
+++ b/ext/json/ext/parser/parser.rl
@@ -95,7 +95,7 @@ static ID i_json_creatable_p, i_json_create, i_create_id, i_create_additions,
           i_chr, i_max_nesting, i_allow_nan, i_symbolize_names,
           i_object_class, i_array_class, i_decimal_class, i_key_p,
           i_deep_const_get, i_match, i_match_string, i_aset, i_aref,
-          i_leftshift, i_new, i_BigDecimal;
+          i_leftshift, i_new, i_BigDecimal, i_freeze, i_uminus;
 
 %%{
     machine JSON_common;
@@ -289,6 +289,10 @@ static char *JSON_parse_value(JSON_Parser *json, char *p, char *pe, VALUE *resul
 
     %% write init;
     %% write exec;
+
+    if (json->freeze) {
+        OBJ_FREEZE(*result);
+    }
 
     if (cs >= JSON_value_first_final) {
         return p;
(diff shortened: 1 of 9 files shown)


casperisfine:
cc @marcandre @tenderlove could I interest one of you into this feature?

casperisfine:
> I presume that a java implementation is needed

I went ahead and implemented it anyway. It wasn't that complicated actually.

ext/json/ext/parser/parser.rl
```diff
@@ -573,7 +577,17 @@ static char *JSON_parse_string(JSON_Parser *json, char *p, char *pe, VALUE *resu
     if (json->symbolize_names && json->parsing_name) {
       *result = rb_str_intern(*result);
     } else if (RB_TYPE_P(*result, T_STRING)) {
+# if STR_UMINUS_DEDUPE_FROZEN
```
hsbt, line 580: @casperisfine Can you add the reason like https://github.com/msgpack/msgpack-ruby/pull/194/files#diff-4bc109d62ab9703449bc38b7290e3cacR164 ?

casperisfine: Sure. However I recently discovered that this code isn't quite correct. I'll push a new version very soon https://github.com/ruby/spec/pull/788

casperisfine: Done.

casperisfine:
It should be good to go now. CI is green.

casperisfine:
@hsbt anything else you'd like changed?

PR #449, opened 2020-10-07
Fix an issue with generate_pretty and empty objects in the Ruby and Java implementations

### LINKED ISSUE #440: Java implementation of JSON.pretty_generate generates different output to MRI for at least an empty hash
From https://github.com/jruby/jruby/issues/6338.

```ruby
require 'json'
puts JSON.pretty_generate({})
```

```
% chruby 2.5.7
% ruby -v test.rb 
ruby 2.5.7p206 (2019-10-01 revision 67816) [x86_64-darwin19]
{
}
% chruby jruby 9.2.11.1
% ruby -v test.rb
jruby 9.2.11.1 (2.5.7) 2020-03-25 b1f55b1a40 OpenJDK 64-Bit Server VM 25.252-b14 on 1.8.0_252-b14 +jit [darwin-x86_64]
{

}
```

### LINKED ISSUE #441: Ruby implementation of JSON.pretty_generate generates different output to C extension for at least an empty hash
From https://github.com/oracle/truffleruby/issues/2053.

```ruby
require 'json'
puts JSON.pretty_generate({})
```

```
% chruby 2.6.5
% ruby -v test.rb
ruby 2.6.5p114 (2019-10-01 revision 67812) [x86_64-darwin19]
{
}
% chruby truffleruby-20.1.0
% ruby -v test.rb
truffleruby 20.1.0, like ruby 2.6.5, GraalVM CE Native [x86_64-darwin]
{

}
```

@eregon then pointed out:

```
$ ruby -v -e 'require "json/pure"; puts JSON.pretty_generate({})'
ruby 2.6.6p146 (2020-03-31 revision 67876) [x86_64-linux]
{

}
```

Possibly related to https://github.com/flori/json/issues/440?

chrisseaton:
Fixed #441.

### DIFF
--- a/java/src/json/ext/Generator.java
+++ b/java/src/json/ext/Generator.java
@@ -334,13 +334,13 @@ public final class Generator {
 
                 buffer.append((byte)'{');
                 buffer.append(objectNl);
-                object.visitAll(new RubyHash.Visitor() {
-                    private boolean firstPair = true;
 
+                final boolean[] firstPair = new boolean[]{true};
+                object.visitAll(new RubyHash.Visitor() {
                     @Override
                     public void visit(IRubyObject key, IRubyObject value) {
-                        if (firstPair) {
-                            firstPair = false;
+                        if (firstPair[0]) {
+                            firstPair[0] = false;
                         } else {
                             buffer.append((byte)',');
                             buffer.append(objectNl);
@@ -360,7 +360,7 @@ public final class Generator {
                     }
                 });
                 state.decreaseDepth();
-                if (objectNl.length() != 0) {
+                if (!firstPair[0] && objectNl.length() != 0) {
                     buffer.append(objectNl);
                     buffer.append(Utils.repeat(state.getIndent(), state.getDepth()));
                 }
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -332,8 +332,10 @@ module JSON
               first = false
             }
             depth = state.depth -= 1
-            result << state.object_nl
-            result << state.indent * depth if indent
+            unless first
+              result << state.object_nl
+              result << state.indent * depth if indent
+            end
             result << '}'
             result
           end
--- a/tests/json_generator_test.rb
+++ b/tests/json_generator_test.rb
@@ -92,6 +92,11 @@ EOT
   end
 
   def test_generate_pretty
+    json = pretty_generate({})
+    assert_equal(<<'EOT'.chomp, json)
+{
+}
+EOT
     json = pretty_generate(@hash)
     # hashes aren't (insertion) ordered on every ruby implementation
     # assert_equal(@json3, json)


chrisseaton:
Also fixes #440 now.

lib/json/pure/generator.rb
```diff
@@ -330,10 +331,13 @@ def json_transform(state)
                 result << %{"#{String(value)}"}
               end
               first = false
+              count += 1
             }
             depth = state.depth -= 1
-            result << state.object_nl
-            result << state.indent * depth if indent
+            unless count.zero?
```
eregon, line 337: Could we use `unless first` to handle this instead of adding `count`?

chrisseaton: I wasn't quite sure about what the 'object' could be and wether it was possibly an enumerator that can only run once?

eregon: `first == true` is always equivalent to `count == 0`, isn't it? (they're modified together in the current diff)

chrisseaton: Ah I see what you mean! Yes I'll do that.

chrisseaton: Done - I thought you were suggesting I called `.first` on the contents.

eregon (approved):
Looks good to me.
@hsbt Could you review and merge?

headius:
It would be helpful to give a JRuby contributor write access to this repository, if this is where we will continue to manage the json extension. That person should probably be me.

java/src/json/ext/Generator.java
```diff
@@ -334,13 +334,13 @@ void generate(final Session session, RubyHash object,
 
                 buffer.append((byte)'{');
                 buffer.append(objectNl);
-                object.visitAll(new RubyHash.Visitor() {
-                    private boolean firstPair = true;
 
+                final boolean[] firstPair = new boolean[]{true};
+                object.visitAll(new RubyHash.Visitor() {
```
headius, line 339: This is fine for the patch, but it's now constructing two objects instead of one. Making this an inner class with an accessible firstPair field would get it back down to one object.

headius (approved):

java/src/json/ext/Generator.java
chrisseaton: Ah, read it out of the object field, yeah.

chrisseaton: It turned out to be quite a lot of refactoring to make the field accessible, as the local holding the visitor needs a new named class or interface in order to be able to read a field out of it, so I didn't do this change. I don't know if you want to @headius? Otherwise I'll leave it as is.

headius:
I believe this has not been released yet... so I am requesting a release! We were unable to resolve https://github.com/jruby/jruby/issues/6338 in JRuby 9.2.14.0 because we do not ship prerelease code for default gems.

PR #470, opened 2021-05-20
Stop including the parser source __LINE__ in exceptions

casperisfine:
It makes testing for JSON errors very tedious. You either have to use a Regexp or to regularly update all your assertions when JSON is upgraded.

```
--- expected
+++ actual
@@ -1 +1 @@
-"is invalid JSON: 809: unexpected token at 'FAIL'"
+"is invalid JSON: 859: unexpected token at 'FAIL'"
```

This is what happened to me today and prompted this PR.

cc @nurse @hsbt @tenderlove


casperisfine:
Sorry for the noise, but could this PR be considered?

PR #488, opened 2021-12-01
Remove unknown keyword arg from DateTime.parse

headius:
This snuck in while addding tests for the `create_additions`
feature. Caught by JRuby when we added the `limit` option to the
Date/DateTime parsing methods, which causes this to be rejected as
an unknown keyword.

### DIFF
--- a/tests/json_addition_test.rb
+++ b/tests/json_addition_test.rb
@@ -165,7 +165,7 @@ class JSONAdditionTest < Test::Unit::TestCase
 
   def test_utc_datetime
     now = Time.now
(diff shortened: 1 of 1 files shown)


headius:
We opted to exclude this test in https://github.com/jruby/jruby/commit/a9f3b6beab0afc0baa82003a83075f3445692b15 because it started to fail when we added the `limit` option. JRuby's date library is still partially in Ruby, and adding the `limit` option there meant it started to reject unknown keywords, like `create_additions.

This was not found by CRuby because the option processing in CRuby's date library does not reject unknown keywords.

PR #491, opened 2022-01-15
Actions workflow - Add new OS's, Ruby 3.1, Ruby master, Windows

MSP-Greg:


MSP-Greg:
This PR uses MSP-Greg/setup-ruby-pkgs, as ragel needs to be installed on all platforms.  Cross platform package installation is essentially the only reason for the action.  It uses the code from ruby/setup-ruby for all the operations it performs.

The PR also adds all MRI head builds, including Windows mswin.  Note that mswin cannot be built on the windows-2022 image due to a breaking change in the C tools in Visual Studio 2022.  MSFT is a little slow at reacting to the issue...

MSP-Greg:
Rebased and updated - ubuntu-22.04, macos-12, Windows mswin

hsbt:
@MSP-Greg Sorry to request again. Can you rebase from master branch? I'd like to merge `ruby/setup-ruby-pkgs@v1` and `ragel` installation changes.

MSP-Greg:
@hsbt Done.  I added `workflow_dispatch`, which can be helpful here or in forks for checking intermittent failures/errors, etc.

MSP-Greg:
@hsbt A little off topic, but would a workflow that checks whether `ext/json/ext/parser/parser.rl` matches `ext/json/ext/parser/parser.c` be helpful?  Puma has one, and it checks that any change to the `*.rl` is also done to the `*.c` file.

Or, it checks that ragel produces the commited c file from the commited rl file...

Regardless, it will be tomorrow at the earliest, 23:35 here.

hsbt:
diff checker of ragel is helpful. But https://github.com/puma/puma/blob/master/.github/workflows/ragel.yml seems too large for only checking. It may be hard to maintain in the future.

PR #492, opened 2022-02-10
Relax Pure::Parser's comment regex to allow any characters in a multi-line comment

sjahu:
The regex used by `Pure::Parser` for ignoring comments in JSON tries to do some fancy handling to prohibit nested comments, but it causes it to choke on strings that would otherwise be legit (i.e. that can be parsed by `Ext::Parser` or by a C compiler).

For example, a couple strings that are parsed correctly by `Ext::Parser` but not `Pure::Parser`:
image

This is the original regex (visualized by [Regexper](https://regexper.com/#%28%3F%3A%5C%2F%5C%2F%5B%5E%5Cn%5Cr%5D*%5B%5Cn%5Cr%5D%7C%5C%2F%5C*%28%3F%3A%5B%5E*%5C%2F%5D%7C%5C%2F%5B%5E*%5D%7C%5C*%5B%5E%5C%2F%5D%7C%5C%2F%28%3F%3D%5C*%5C%2F%29%29*%5C*%5C%2F%7C%5B%20%5Ct%5Cr%5Cn%5D%2B%29%2B)):
image

This is the simplified regex after my changes ([Regexper](https://regexper.com/#%28%3F%3A%5C%2F%5C%2F%5B%5E%5Cn%5Cr%5D*%5B%5Cn%5Cr%5D%7C%5C%2F%5C*%28%3F%3A%5B%5Cs%5CS%5D*%3F%29*%5C*%5C%2F%7C%5B%20%5Ct%5Cr%5Cn%5D%2B%29%2B)):
image

I added a test to assert that a comment containing `/*` is parseable. Note that the test case immediately above that change asserts that a nested ANSI C-style comment still triggers an exception.

### DIFF
--- a/lib/json/pure/parser.rb
+++ b/lib/json/pure/parser.rb
@@ -39,10 +39,7 @@ module JSON
          //[^\n\r]*[\n\r]| # line comments
          /\*               # c-style comments
          (?:
-          [^*/]|        # normal chars
-          /[^*]|        # slashes that do not start a nested comment
-          \*[^/]|       # asterisks that do not end this comment
-          /(?=\*/)      # single slash before this comment's end
+          [\s\S]*?         # any char, repeated lazily
          )*
            \*/               # the End of this comment
            |[ \t\r\n]+       # whitespaces: space, horizontal tab, lf, cr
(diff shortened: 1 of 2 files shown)


sjahu:
(CI hasn't run because "First-time contributors need a maintainer to approve running workflows.")

PR #494, opened 2022-03-13
Make rb_scan_args handle keywords more similar to Ruby methods (#2460)

maysam:
Making it compatibile with Ruby 2.7.5

Actually just fixing the deprecation notice


hsbt:
Unfortunately, v1.8 is EOL status.

Issue #503, opened 2022-09-12
IPAddr#to_json remove prefix address with CIDR


taketo1113:
IPAddr#to_json removed the prefix address with IPAddr/CIDR value.
It unable to deserialize IPAddr/CIDR value.

### Expected behavior
```ruby
require 'date'
require 'json'

IPAddr.new('172.16.0.0/24').to_json #=> "\"172.16.0.0/24\""
```

### Actual behavior
```ruby
require 'ipaddr'
require 'json'

IPAddr.new('172.16.0.0/24').to_json #=> "\"172.16.0.0\""
```

### Related URL
- https://github.com/rails/rails/issues/46006

byroot:
The JSON gem has no specific support for `IpAddr`. I suspect this is a change in `IpAddr#to_s` which is what JSON call when it doesn't know about a particular type.

taketo1113:
Thanks to reply.

> The JSON gem has no specific support for IpAddr. 

I will create PR to add `IPAddr#to_json` method in ipaddr gem.
Becase I think the changing behavior of `IPAddr#to_s` has breaking compatibility.

see. https://github.com/rails/rails/pull/52365

taketo1113:
cc: @byroot 
I have created PR to add `IPAddr#to_json/as_json` for ipaddr gem.
https://github.com/ruby/ipaddr/pull/77

PR #504, opened 2022-09-12
Add support IPAddr class

taketo1113:
Add support IPAddr class to deserialize IPAddr/CIDR that ip address with prefix address.

- #503


byroot:
Thanks for the PR but I don't wish to add any more `json/add` classes.

taketo1113:
Thanks to reply. I got it.

hsbt:
FYI: https://bugs.ruby-lang.org/issues/19528

PR #506, opened 2022-11-10
tests/ractor_test.rb: make assert_separately available

lucaskanashiro:
Require tests/lib/helper.rb to avoid:

NoMethodError: undefined method `assert_separately'

This issue has been seen in Debian and Ubuntu and fixed with the proposed patch.


hsbt:
How do you run test suite? I couldn't reproduce with `rake test`.

PR #507, opened 2023-01-12
Fix "unexpected token" offset for Infinity

jhawthorn:
Previously in the JSON::Ext parser, when we encountered an "Infinity" token (and we weren't allowing `NaN`/`Infinity`) we would try to display the "unexpected token" at the character before.

This was found by OSS-Fuzz due to its buffer underflow of exactly 1 byte. Due to low impact this isn't being tracked as a security issue.


hsbt:
@jhawthorn Thanks! Can you omit or pend with failing tests like `RUBY_PLATFORM =~ /java/`?

PR #513, opened 2023-01-29
use OBJ_FREEZE_RAW for string keys

luke-gru:
This obviates the need to transition the string to a new shape, which means there doesn't have to be a global VM lock, which hurts performance when using JSON.parse in ractors.


luke-gru:
On second thought, even though this works right now I don't know what the guarantees are for OBJ_FREEZE_RAW in the future, so I'm going to close this.

PR #516, opened 2023-02-08
JRuby requires a minimum of Java 8

headius:
More recent JDKs will reject this command line because 1.6 is too old to build for.


headius:
Compile errors are fixed in my other PR, #517.

hsbt:
@headius Can you rebase from master branch?

headius:
Will do!

PR #518, opened 2023-02-20
Use ruby_xfree to free buffers

casperisfine:
They are allocated with ruby_xmalloc, they should be freed with ruby_xfree.

cc @luke-gru


hsbt:
@casperisfine Sorry, late response. It seems fine.

PR #519, opened 2023-02-20
Add a `strict` option to Generator

casperisfine:
The default behavior of calling `to_json` or even `to_s` on unsupported object is problematic in many cases.

```
>> JSON.parse(JSON.dump(Object.new))
=> "#<Object:0x00000001034d8420>"
```

In some cases like blind serailization into some kind of store, you may want to assert that `JSON.parse(JSON.dump(obj)) == obj`.

The `strict` option is a way to ensure this by rejecting any type that doesn't map cleanly to a JSON native type.

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -131,6 +131,7 @@ module JSON
           @allow_nan             = false
           @ascii_only            = false
           @escape_slash          = false
+          @strict                = false
           @buffer_initial_length = 1024
           configure opts
         end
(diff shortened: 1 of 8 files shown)


gjtorikian:
@flori @hsbt Would it be possible to have this reviewed?

hsbt:
@casperisfine Thank you. I picked your commit and fixed conflict with https://github.com/flori/json/pull/525

lib/json/pure/generator.rb
```diff
@@ -324,7 +341,9 @@ def json_transform(state)
               result << state.space_before
               result << ':'
               result << state.space
-              if value.respond_to?(:to_json)
+              if state.strict?
+                raise GeneratorError, "#{value.class} not allowed in JSON"
+              elsif value.respond_to?(:to_json)
```
eregon, line 346: This is a bug, see https://github.com/flori/json/issues/584

PR #521, opened 2023-03-08
Update CI matrix

hsbt:


hsbt:
@headius JRuby 9.4 and head are failed on Windows https://github.com/flori/json/actions/runs/4360822114/jobs/7624157859

headius:
It looks like a problem building the extension. The classpath is not set up properly to point at JRuby, it seems...

```
javac -classpath java/src:D:/jruby-9.4.1.0/lib/jruby.jar -source 1.6 -target 1.6 java/src/json/ext/ByteListTranscoder.java
warning: [options] bootstrap class path not set in conjunction with -source 1.6
java\src\json\ext\ByteListTranscoder.java:8: error: package org.jruby.exceptions does not exist
import org.jruby.exceptions.RaiseException;
                           ^
java\src\json\ext\ByteListTranscoder.java:9: error: package org.jruby.runtime does not exist
import org.jruby.runtime.ThreadContext;
                        ^
java\src\json\ext\ByteListTranscoder.java:10: error: package org.jruby.util does not exist
import org.jruby.util.ByteList;
                     ^
...
```

Someone will have to give this a try on Windows. I'm guessing this has never built properly because we usually don't need to build the extension more than once on one platform for release.

PR #525, opened 2023-04-14
Rename escape_slash in script_safe and also escape E+2028 and E+2029

casperisfine:
Fix: https://github.com/flori/json/pull/215
Fix:https://github.com/flori/json/issues/214

It is rather common to directly interpolate JSON string inside <script> tags in HTML as to provide configuration or parameters to a script.

However this may lead to XSS vulnerabilities, to prevent that 3 characters need to be escaped:

  - `/` (forward slash)
  - `U+2028` (LINE SEPARATOR)
  - `U+2029` (PARAGRAPH SEPARATOR)

The forward slash need to be escaped to prevent closing the script tag early, and the other two are valid JSON but invalid Javascript and can be used to break JS parsing.

Given that the intent of escaping forward slash is the same than escaping U+2028 and U+2029, I chose to rename and repurpose the existing `escape_slash` option.

cc @hsbt @nurse 

This could be used to very [significantly speedup `ActiveSupport::JSON`](https://github.com/rails/rails/blob/444df0eee1b537ecaa11509e819b071d4e87b519/activesupport/lib/active_support/json/encoding.rb#L42-L54) (cc @jhawthorn)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Changes
 
+* `escape_slash` option was renamed as `script_safe` and now also escape U+2028 and U+2029. `escape_slash` is now an alias of `script_safe`.
+
 ### 2021-10-24 (2.6.1)
 
 * Restore version.rb with 2.6.1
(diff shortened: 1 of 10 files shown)


hsbt:
@casperisfine Can you also update https://github.com/flori/json/blob/master/Rakefile#L171?

casperisfine:
@hsbt done. Also sorry for including this change, I'm not sure what was going on but I couldn't get JRuby to work with that `json-java.gemspec`.

But I can revert that part, no problem.

casperisfine:
Some note here.

As implemented in this PR, this option make the json safe to include in `<script>` tags, but not elsewhere in HTML.

Some JSON libraries do offer an "xss_safe" mode, that also escape `<` as `\u003e`. It may make sense to implement it like this here, as to make the option more generally safe.

casperisfine:
Ok, I reverted the Gemfile and gemspec changes because it was failing on CI. I don't know what's wrong with JRuby on my machine :/

PR #526, opened 2023-05-26
Improvements for `decimal_class: BigDecimal`

notEthan:
I think it is excellent that it is so easy to preserve arbitrary numeric precision parsing JSON with `decimal_class: BigDecimal`. This currently has a couple of deficiencies, though.

- Not round-trippable: currently JSON.generate will represent BigDecimal as a string. I'm not sure why this is, when it can be encoded perfectly as a number.
- Obscure documentation - it's not mentioned on the ::JSON module or on the default (ext) parser. It is mentioned on the pure and java parsers.

This PR adds documentation where missing, and changes the generation to be a number, as seems much more natural.

I also split `json_addition_test.rb` to its own run because `require 'json/add/bigdecimal'` would override `BigDecimal#to_json`.


notEthan:
The implementation for c ext is in progress, and I'm not sure about implementing a java version. I wanted to put this up as a draft to solicit feedback while I address those things.

notEthan:
I added a c ext implementation but I am very inexperienced with c ruby extensions and it could probably be better. I will wait and see about feedback before I try to improve it further or even think about java.

byroot:
I think this had too many incompatibility concerns. I'd recommend just defining a custom `decimal_class` that behave as you desire.

notEthan:
Amazing feedback a year and a half later. Even if there are many incompatibility concerns (you think, who knows) the documentation is still very lacking, though my experience here sure doesn't have me inclined to try (re)submitting any other improvement.

PR #537, opened 2023-08-14
Always dup argument to preserve original encoding for force_encoding

### LINKED ISSUE #529: JSON.parse Alters Encoding of Passed String Unexpectedly
### Description

I am currently developing an application where string encoding is important, and I've discovered that JSON.parse unexpectedly alters the encoding of the passed string.

Here's a simple code snippet to reproduce the issue:

```ruby
require 'json'

str = '{ "hoge": "fuga" }'.force_encoding('ASCII-8BIT')

puts 'JSON'
puts "Before: #{str.encoding}"
puts JSON.parse(str)
puts "After:  #{str.encoding}"

puts

str = '{ "hoge": "fuga" }'.force_encoding('ASCII-8BIT')

puts 'JSON::Ext::Parser'
puts "Before: #{str.encoding}"
puts JSON::Ext::Parser.new(str).parse
puts "After:  #{str.encoding}"
```

This code yields the following output:

```
JSON
Before: ASCII-8BIT
{"hoge"=>"fuga"}
After:  UTF-8

JSON::Ext::Parser
Before: ASCII-8BIT
{"hoge"=>"fuga"}
After:  UTF-8
```

As shown above, the encoding of the string str changes from 'ASCII-8BIT' to 'UTF-8' after the JSON.parse operation.

My current environment is as follows:

```
$ ruby --version
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-darwin22]
$ ruby -rjson -e 'puts JSON::VERSION'
2.6.3
```

## Expected Behavior

I would expect JSON.parse not to alter the original encoding of the passed string, unless it is necessary for the parsing operation.

## Current Workaround

Currently, I'm using the dup method to create a copy of the original string before parsing it to prevent the original string's encoding from being altered:

```ruby
JSON.parse(str.dup)
```

aeroastro:
This fixes #529.

In order to avoid changing original encoding,
this patch always dups argument for `ASCII-8BIT` encoding.

I have also added a test `test_argument_encoding_for_binary`.
The structure is following `test_argument_encoding`.

Considering the following facts, IMHO, I believe this patch meets the original intention of this gem.

* `JSON::Pure::Parser.new(str)` does not change the encoding of original argument for `ASCII-8BIT`
* There already exists a test to guarantee that the encoding of argument is not changed for non `ASCII-8BIT` encodings.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -1778,10 +1778,7 @@ static VALUE convert_encoding(VALUE source)
   }
 
  if (encindex == binary_encindex) {
-    if (OBJ_FROZEN(source)) {
-      source = rb_str_dup(source);
-    }
-    return rb_enc_associate_index(source, utf8_encindex);
+    return rb_enc_associate_index(rb_str_dup(source), utf8_encindex);
   }
 
   return rb_str_conv_enc(source, rb_enc_from_index(encindex), rb_utf8_encoding());
@@ -1912,7 +1909,7 @@ static VALUE cParser_initialize(int argc, VALUE *argv, VALUE self)
 }
 
 
-#line 1916 "parser.c"
+#line 1913 "parser.c"
 enum {JSON_start = 1};
 enum {JSON_first_final = 10};
 enum {JSON_error = 0};
@@ -1920,7 +1917,7 @@ enum {JSON_error = 0};
 enum {JSON_en_main = 1};
 
 
-#line 824 "parser.rl"
+#line 821 "parser.rl"
 
 
 /*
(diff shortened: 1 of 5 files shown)


Dani-beef:
require 'json'

str = '{ "hoge": "fuga" }'.force_encoding('ASCII-8BIT')

puts 'JSON'
puts "Before: #{str.encoding}"
puts JSON.parse(str)
puts "After:  #{str.encoding}"

puts

str = '{ "hoge": "fuga" }'.force_encoding('ASCII-8BIT')

puts 'JSON::Ext::Parser'
puts "Before: #{str.encoding}"
puts JSON::Ext::Parser.new(str).parse
puts "After:  #{str.encoding}"

byroot:
Thanks. I rebased your PR and also added the Java version of the fix.

PR #539, opened 2023-09-05
Use test-unit-ruby-core gem

hsbt:


hsbt:
We should consider to bump major version because `json` will drop to suppot < Ruby 2.5.

PR #543, opened 2023-10-09
Do not use BigDecimal.new to construct

headius:
Fixes jruby/jruby#7752

### DIFF
--- a/java/src/json/ext/Parser.java
+++ b/java/src/json/ext/Parser.java
@@ -1195,7 +1195,7 @@ case 5:
             Ruby runtime = getRuntime();
             ByteList num = absSubSequence(p, new_p);
             IRubyObject numString = runtime.newString(num.toString());
-            return parser.decimalClass.callMethod(context, "new", numString);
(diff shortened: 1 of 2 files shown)


hsbt:
@headius I fixed CI failure with https://github.com/flori/json/pull/544. Can you rebase this from HEAD?

headius:
@hsbt Done!

kares:
while this is :+1:, it's also breaking in terms of the `decimal_class` feature, which is an option `JSON` still supports.

kares:
building on the changes here, I've come up with: https://github.com/flori/json/pull/546 ... which should be "more" :back:wards compatible

hsbt:
@headius @kares I have no idea to merge this or #546.

kares:
I would obviously vote for #546 but let's wait for @headius to do a review. 

The difference between the two is that this PR will not consider a user provided custom decimal_class option, while #546 will.

headius:
#546 should be merged as it includes these fixes while presenting backward compatibility. Thanks @kares !

PR #546, opened 2023-10-15
[jruby] avoid using deprecated BigDecimal.new

kares:
an alternative PR to: https://github.com/flori/json/pull/543 - for the `BigDecimal` case it should behave the same

main difference is that a custom `opts[:decimal_class]` (e.g. a `BigDecimal` sub-class) should work as before... 

not sure `decimal_class` feature is used much or that folks would set it to anything different than `BigDecimal` but at least the pure.rb seems to [retain a similar logic](https://github.com/flori/json/blob/766f816/lib/json/pure/parser.rb#L214-L222)


headius:
Looking forward to a release of this so we can close the relevant bugs in JRuby: https://github.com/jruby/jruby/issues/7752

PR #554, opened 2023-12-01
JSON.dump: handle unenclosed hashes regression

casperisfine:
Fix: https://github.com/flori/json/issues/553

We can never add keyword arguments to `dump` otherwise existing code using unenclosed hash will break.

cc @hsbt

### DIFF
--- a/lib/json/common.rb
+++ b/lib/json/common.rb
@@ -611,7 +611,7 @@ module JSON
   #   puts File.read(path)
   # Output:
   #   {"foo":[0,1],"bar":{"baz":2,"bat":3},"bam":"bad"}
-  def dump(obj, anIO = nil, limit = nil, strict: NOT_SET)
+  def dump(obj, anIO = nil, limit = nil, kwargs = nil)
     if anIO and limit.nil?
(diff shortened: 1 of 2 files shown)


hsbt:
👍

mame:
@byroot Now `JSON.dump({}, strict: true)` would not work well. Is it ok?

byroot:
No it's not... but since it's a new feature it's less if a problem. I'll submit another PR soon. Thank you for letting me know

k0kubun:
If I may: https://github.com/flori/json/pull/556

lib/json/common.rb
```diff
@@ -637,6 +637,12 @@ def dump(obj, anIO = nil, limit = nil, strict: NOT_SET)
   def self.iconv(to, from, string)
     string.encode(to, from)
   end
+
+  private
+
+  def merge_dump_options(opts, strict: NOT_SET)
+    opts[:strict] = strict if NOT_SET != strict
```
k0kubun, line 644: This can destructively modify `JSON.dump_default_options`. I addressed that in https://github.com/flori/json/pull/556.

PR #558, opened 2023-12-04
Fix JSON.dump overload combination

tompng:
Fixed this overload
```ruby
# wrong argument type Hash (expected Integer)
dump({}, io, strict: true)
dump({}, limit, strict: true)
```

And fixed this NoMethodError
```ruby
# Works and tested in test code
dump({}, strict: true) # => "{}"
# NoMethodError. was not tested in test code
JSON.dump({}, strict: true) # => undefined method `merge_dump_options' for JSON:Module
```

### DIFF
--- a/lib/json/common.rb
+++ b/lib/json/common.rb
@@ -612,16 +612,13 @@ module JSON
   # Output:
   #   {"foo":[0,1],"bar":{"baz":2,"bat":3},"bam":"bad"}
   def dump(obj, anIO = nil, limit = nil, kwargs = nil)
-    if anIO and limit.nil?
-      anIO = anIO.to_io if anIO.respond_to?(:to_io)
-      unless anIO.respond_to?(:write)
-        if kwargs.nil? and anIO.is_a?(Hash)
-          kwargs = anIO
-        else
-          limit = anIO
-        end
-        anIO = nil
-      end
+    io_limit_opt = [anIO, limit, kwargs].compact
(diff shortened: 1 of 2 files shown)


lib/json/common.rb
```diff
@@ -642,12 +639,14 @@ def self.iconv(to, from, string)
     string.encode(to, from)
   end
 
-  private
```
tompng, line 645: `merge_dump_options` should be private instance method and also private module function
```ruby
module_function

def a_private_instance_method_and_public_module_function
end

private # module_function rule is cleared

def a_private_instance_method # not a module function
end
```

lib/json/common.rb
```diff
@@ -612,16 +612,13 @@ class << self
...
-        end
-        anIO = nil
-      end
+    io_limit_opt = [anIO, limit, kwargs].compact
+    kwargs = io_limit_opt.pop if io_limit_opt.last.is_a?(Hash)
+    anIO, limit = io_limit_opt
+    if anIO.respond_to?(:to_io)
+      anIO = anIO.to_io
+    elsif limit.nil? && !anIO.respond_to?(:write)
+      anIO, limit = nil, anIO
```
tompng, line 621: Fixed the inconsistency
```ruby
dump(obj, io, nil) #=> io.to_io was called
dump(obj, io, 10) #=> io.to_io was not called
```

And if `to_io` is called, the result of `to_io` should be an IO, no need to check for `respond_to?(:write)`

PR #561, opened 2023-12-05
Use rb_sym2str instead of SYM2ID

jhawthorn:
This avoids pinning an id to the symbol used if a dynamic symbol is passed in as a hash key.

rb_sym2str is available in Ruby 2.2+ and json depends on >= 2.3.


hsbt:
👍

PR #565, opened 2024-01-31
Make OpenStruct support as optional

hsbt:
see https://github.com/ruby/ostruct/blob/master/lib/ostruct.rb#L136

`OpenStruct` is not recommend to use with performance reason today. We should make `OpenStruct` support as optional.


Earlopain:
Would this not warn anyways? The `logging` gem for example requires `syslog` in rescue like in this PR but even though the functionality is optional it still logs the message.

PR #572, opened 2024-03-09
Update README.md

mthirumalai2905:
added a link which generates upto 1000 fake json datas within seconds


hsbt:
We don't need to recommend this site officially. It's easy to create dummy data as JSON.

PR #576, opened 2024-03-26
Remove references to "19" methods in JRuby

headius:
These methods have been deprecated for some time and we would like to delete them.

See jruby/jruby#9143 and jruby/jruby#8149.

### DIFF
--- a/java/src/json/ext/Parser.java
+++ b/java/src/json/ext/Parser.java
@@ -278,7 +278,7 @@ public class Parser extends RubyObject {
     }
 
     private RubyFloat createFloat(final ThreadContext context, final ByteList num) {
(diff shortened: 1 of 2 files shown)


headius:
The `byteListToDouble19` equivalent is now `byteListToDouble`, but rather than backport that I have used the implementation it contains.

headius:
This appears to be passing on all JRuby versions back to 9.1, and can be released any time.

We will want this in the wild for a while before we can proceed with removing the 19 versions, which mostly will happen in JRuby 10 (later this year) and a subsequent major release.

headius:
@hsbt Thank you for merging! We can release this any time.

PR #580, opened 2024-04-15
Add a simple JSON.dump and JSON.load benchmark

### LINKED ISSUE #117: Don't include benchmark results in gem
AFAIK there's no reason to include the benchmark results in the gem. But it adds 1.1M to 1.4M of sensible data. In today's world, this seems like nothing, but it adds up when having many versions and multiple locations where they can be.

Btw: a big thank you for this cornerstone of the ruby programming universe!

eregon:
* Restoring benchmarks/ohai.json from dad9fbfbfbfcfe743563348f2c17aedecfabe803.

The commit intended to fix #117, that's still fixed as these new files are not included in the gem `s.files =`.

### DIFF
--- a/benchmarks/bench.rb
+++ b/benchmarks/bench.rb
@@ -1,0 +1,41 @@
+require 'benchmark/ips'
+
+$:.unshift File.expand_path('../ext', __dir__)
+$:.unshift File.expand_path('../lib', __dir__)
+
+bench, mode = ARGV
+
+if mode == 'pure'
+  require 'json/pure'
+else
+  require 'json/ext'
+end
+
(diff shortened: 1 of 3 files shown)


eregon:
@hsbt Could you review this too? (if not, who should I ask?)

benchmarks/bench.rb
```diff
@@ -0,0 +1,41 @@
+require 'benchmark/ips'
```
hsbt, line 1: Can you add `benchmark-ips` to Gemfile?

eregon: Done

hsbt:
👍

PR #585, opened 2024-05-09
JSON::Pure fix strict mode

casperisfine:
Followup: https://github.com/flori/json/pull/519
Fix: https://github.com/flori/json/issues/584

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -219,7 +219,9 @@ module JSON
           @script_safe
         end
 
-        # Returns true, if forward slashes are escaped. Otherwise returns false.
+        # Returns true, if strict mode is enabled. Otherwise returns false.
+        # Strict mode only allow serializing JSON native types: Hash, Array,
+        # String, Integer, Float, true, false and nil.
         def strict?
           @strict
         end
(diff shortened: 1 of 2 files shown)


lib/json/pure/generator.rb
```diff
@@ -354,7 +354,7 @@ def json_transform(state)
               result << delim unless first
               result << state.indent * depth if indent
               result = "#{result}#{key.to_s.to_json(state)}#{state.space_before}:#{state.space}"
-              if state.strict?
+              if state.strict? && !(!value || value == true || value.is_a?(Array) || value.is_a?(Integer) || value.is_a?(Float) || value.is_a?(String))
```
eregon, line 357: This looks expensive, can't we just rely on `.to_json(state)` calling e.g. Object#to_json and that will raise?

eregon:
I also noticed the docs in https://github.com/flori/json/blob/4f876a8bf7c45972800bae5bdbaf892aa2f714a5/lib/json/pure/generator.rb#L222 are wrong, could you fix that too?

And the alignment in https://github.com/flori/json/blob/4f876a8bf7c45972800bae5bdbaf892aa2f714a5/lib/json/pure/generator.rb#L143 is inconsistent (I dislike this kind of alignment, too high maintenance and diff noise, but might as well be consistent in one way, i.e. all or none aligned)

lib/json/pure/generator.rb
casperisfine: I don't think we can as it wouldn't catch objects with a custom `to_json`.

casperisfine:
> I also noticed

I'd rather not bundle multiple unrelated fixes in the same PR. Makes merging more complicated.

eregon:
Ah sorry, it's `@script_safe` misaligned, and that's unrelated indeed.

OTOH I think fixing the description of `def strict?` is very related.

lib/json/pure/generator.rb
eregon: Re `!value` that could be an issue if someone redefines `!` on their type to return true in some cases.
Not sure whether it's worth caring about, redefining `!` seems a pretty poor idea (maybe it should be a Ruby warning even).

eregon (approved):
Thanks for the fix!

PR #586, opened 2024-05-09
Optimize JSON::Pure::Generator by 2x-4x for simple options cases

eregon:
```
* ruby --yjit benchmarks/bench.rb dump pure ruby 3.3.1 (2024-04-23 revision c56cd86388) +YJIT [x86_64-linux]
  Before: JSON.dump(obj)    604.604  (± 0.3%) i/s    (1.65 ms/i) -      3.060k in   5.061200s
  After:  JSON.dump(obj)      2.531k (± 0.4%) i/s  (395.14 μs/i) -     12.801k in   5.058326s
* ruby benchmarks/bench.rb dump pure
  truffleruby 24.1.0-dev-a8ebb51b, like ruby 3.2.2, Oracle GraalVM JVM [x86_64-linux]
  Before: JSON.dump(obj)      3.728k (± 9.4%) i/s  (268.26 μs/i) -     18.559k in   5.068915s
  After:  JSON.dump(obj)      7.835k (± 8.5%) i/s  (127.63 μs/i) -     39.004k in   5.031116s
```
Using the benchmark from https://github.com/flori/json/pull/580

BTW, the C extension, with YJIT gives:
```
JSON::Ext::Generator
ruby 3.3.1 (2024-04-23 revision c56cd86388) +YJIT [x86_64-linux]
JSON.dump(obj)      7.402k (± 1.0%) i/s  (135.10 μs/i) -     37.485k in   5.064832s
```

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -239,6 +239,8 @@ module JSON
           opts.each do |key, value|
             instance_variable_set "@#{key}", value
           end
+
+          # NOTE: If adding new instance variables here, check whether #generate should check them for #generate_json
           @indent                = opts[:indent] if opts.key?(:indent)
           @space                 = opts[:space] if opts.key?(:space)
           @space_before          = opts[:space_before] if opts.key?(:space_before)
@@ -288,10 +290,68 @@ module JSON
         # created this method raises a
         # GeneratorError exception.
         def generate(obj)
-          result = obj.to_json(self)
+          if @indent.empty? and @space.empty? and @space_before.empty? and @object_nl.empty? and @array_nl.empty? and
+              !@ascii_only and !@script_safe and @max_nesting == 0 and !@strict
+            result = generate_json(obj, '')
+          else
+            result = obj.to_json(self)
+          end
           JSON.valid_utf8?(result) or raise GeneratorError,
             "source sequence #{result.inspect} is illegal/malformed utf-8"
           result
+        end
+
+        # Handles @allow_nan, @buffer_initial_length, other ivars must be the default value (see above)
+        private def generate_json(obj, buf)
+          case obj
+          when Hash
+            buf << '{'.freeze
+            first = true
+            obj.each_pair do |k,v|
+              buf << ','.freeze unless first
+              fast_serialize_string(k.to_s, buf)
+              buf << ':'.freeze
+              generate_json(v, buf)
+              first = false
+            end
+            buf << '}'.freeze
+          when Array
+            buf << '['.freeze
+            first = true
+            obj.each do |e|
+              buf << ','.freeze unless first
+              generate_json(e, buf)
+              first = false
+            end
+            buf << ']'.freeze
+          when String
+            fast_serialize_string(obj, buf)
+          when Integer
+            buf << obj.to_s
+          else
+            # Note: Float is handled this way since Float#to_s is slow anyway
+            buf << obj.to_json(self)
+          end
+        end
+
+        # Assumes !@ascii_only, !@script_safe
+        if Regexp.method_defined?(:match?)
+          private def fast_serialize_string(string, buf) # :nodoc:
+            buf << '"'.freeze
+            string = string.encode(::Encoding::UTF_8) unless string.encoding == ::Encoding::UTF_8
+
+            if /["\\\x0-\x1f]/n.match?(string)
+              buf << string.gsub(/["\\\x0-\x1f]/n, MAP)
+            else
+              buf << string
+            end
+            buf << '"'.freeze
+          end
+        else
+          # Ruby 2.3 compatibility
+          private def fast_serialize_string(string, buf) # :nodoc:
+            buf << string.to_json(self)
+          end
         end
 
         # Return the value returned by method +name+.


eregon:
@hsbt Could you review this as well? :pray: 

---

cc @byroot I mentioned this to you at RubyKaigi.
Might also be interesting to @maximecb ([this](https://github.com/Shopify/yjit-bench/blob/main/benchmarks/ruby-json/benchmark.rb) is a pure-Ruby JSON parser, this PR is optimizing the pure-Ruby JSON generator).

lib/json/pure/generator.rb
```diff
@@ -286,12 +288,70 @@ def to_h
...
             "source sequence #{result.inspect} is illegal/malformed utf-8"
           result
         end
 
+        # Handles @allow_nan, @buffer_initial_length, other ivars must be the default value (see above)
+        private def generate_json(obj, buf)
+          case obj
+          when Hash
+            buf << '{'.freeze
+            first = true
```
maniSHarma7575, line 307: Instead of maintaining the extra variable for the state (`first`). We can use `each_with_index`
```ruby
  when Hash
    buf << '{'
    obj.each_with_index do |(k, v), i|
      buf << ',' unless i.zero?
      fast_serialize_string(k.to_s, buf)
      buf << ':'
      generate_json(v, buf)
    end
    buf << '}'
```

eregon: That's likely much slower as it causes more Array allocations.
Also the code is more similar to Hash#to_json this way.

lib/json/pure/generator.rb
```diff
@@ -286,12 +288,70 @@ def to_h
         # created this method raises a
         # GeneratorError exception.
         def generate(obj)
-          result = obj.to_json(self)
+          if @indent.empty? and @space.empty? and @space_before.empty? and @object_nl.empty? and @array_nl.empty? and
+              !@ascii_only and !@script_safe and @max_nesting == 0 and !@strict
```
byroot, line 292: This feels like this could be precomputed in `configure` so you'd only need to check a single ivar, no?

lib/json/pure/generator.rb
```diff
@@ -286,12 +288,70 @@ def to_h
...
+
+            if /["\\\x0-\x1f]/n.match?(string)
+              buf << string.gsub(/["\\\x0-\x1f]/n, MAP)
+            else
+              buf << string
+            end
+            buf << '"'.freeze
+          end
+        else
+          # Ruby 2.3 compatibility
```
byroot, line 349: At this point we might consider bumping the required version to 2.4 or even 2.5

lib/json/pure/generator.rb
```diff
@@ -286,12 +288,70 @@ def to_h
...
+              generate_json(e, buf)
+              first = false
+            end
+            buf << ']'.freeze
+          when String
+            fast_serialize_string(obj, buf)
+          when Integer
+            buf << obj.to_s
+          else
+            # Note: Float is handled this way since it is Float#to_s is slow anyway
```
byroot, line 330:
```suggestion
            # Note: Float is handled this way since Float#to_s is slow anyway
```

byroot (approved):
Some minor comments, but looks good.

lib/json/pure/generator.rb
eregon: Mmh, I'm not sure what's the protocol, the State class has a bunch of public `attr_accessor` so it looks like `configure` is not necessarily called after such accessors would be used to write.
Or maybe JSON users are not supposed to call these accessors? What are they for then?

BTW it seems pretty inefficient that e.g. JSON.generate would create a new State every time (which runs the State constructor and configure every time) instead of using `SAFE_STATE_PROTOTYPE.dup` or so, but that's a separate thing so improve later.

lib/json/pure/generator.rb
eregon: That would be nice, but this is not much overhead, so that could be cleaned up later when we bump the min Ruby version.

lib/json/pure/generator.rb
eregon: Done

lib/json/pure/generator.rb
eregon: Out of curiosity I did 
```patch
diff --git a/lib/json/pure/generator.rb b/lib/json/pure/generator.rb
index 9f44ef5..d07c2de 100644
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -268,6 +268,10 @@ module JSON
           else
             @max_nesting = 0
           end
+
+          @fast = (@indent.empty? and @space.empty? and @space_before.empty? and @object_nl.empty? and @array_nl.empty? and
+            !@ascii_only and !@script_safe and @max_nesting == 0 and !@strict)
+
           self
         end
         alias merge configure
@@ -277,6 +281,7 @@ module JSON
         def to_h
           result = {}
           instance_variables.each do |iv|
+            next if iv == :@fast
             iv = iv.to_s[1..-1]
             result[iv.to_sym] = self[iv]
           end
@@ -292,8 +297,10 @@ module JSON
         def generate(obj)
           if @indent.empty? and @space.empty? and @space_before.empty? and @object_nl.empty? and @array_nl.empty? and
               !@ascii_only and !@script_safe and @max_nesting == 0 and !@strict
+            raise "OOPS" if !@fast
             result = generate_json(obj, '')
           else
+            raise "OOPS" if @fast
             result = obj.to_json(self)
           end
           JSON.valid_utf8?(result) or raise GeneratorError,
```
To see if there were any inconsistency and the test suite didn't find any.
But still, I think we need to understand better how these accessors are used (or maybe remove them if they're not needed).

eregon:
Rebased and CI should be green

PR #587, opened 2024-05-31
Cleanup useless ifdef

casperisfine:
The json gem now requires Ruby 2.3, so there is no point keeping compatibility code for older releases that don't have the TypedData API.

cc @hsbt


BabydollSacca (commented):
Idk

PR #594, opened 2024-09-02
Remove outdated ifdef checks

casperisfine:
`json` requires Ruby 2.3, so `HAVE_RUBY_ENCODING_H` and `HAVE_RB_ENC_RAISE` are always true.

cc @hsbt


hsbt:
I confirmed this with @nobu

PR #599, opened 2024-09-06
Restore the benchmarks, make it easy to run them

LukeShu:
As pointed out in https://github.com/flori/json/pull/580, it is safe to re-add the benchmarks, as long as they are not included in `s.files`.

This restores the old benchmarks, fixes a some bit-rot in them to get them running again, and adds a semi-automated way to update a nice little summary of the benchmarks in the README.


byroot:
Thanks for the PR, but I went with importing another suite of benchmarks, in https://github.com/ruby/json/pull/606

Issue #344, opened 2017-11-21
String#to_json behavior different for pure vs native ext


nirvdrum:
The `String#to_json` method added when "json" is required has slightly different semantics for the pure vs native extension. In particular, a String with an invalid encoding throws an exception with the native exception but does not with the pure variant.

```
kjmenard@melchior /t/fog-json> ruby -ve 'require "json"; p "\x82\xAC\xEF".to_json'
ruby 2.3.5p376 (2017-09-14 revision 59905) [x86_64-linux]
-e:1:in `to_json': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
	from -e:1:in `<main>'

kjmenard@melchior /t/fog-json> ruby -ve 'require "json/pure"; p "\x82\xAC\xEF".to_json'
ruby 2.3.5p376 (2017-09-14 revision 59905) [x86_64-linux]
"\"\x82\xAC\xEF\""
```

I think the pure version should raise an exception `unless string.valid_encoding?`.

marcandre:
Confirmed with ruby 2.7.1p83

PR #602, opened 2024-10-08
JSONPure: String#to_json should raise on invalid encoding

byroot:
Fix: #344

This matches the ext behavior.

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -450,6 +450,9 @@ module JSON
           def to_json(state = nil, *args)
             state = State.from_state(state)
             if encoding == ::Encoding::UTF_8
+              unless valid_encoding?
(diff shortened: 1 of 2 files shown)


nirvdrum:
Thank you. Now that the gem is seeing active development I went to fix #344 and was pleasantly surprised to see you already address it.

lib/json/pure/generator.rb
```diff
@@ -450,6 +450,9 @@ module String
           def to_json(state = nil, *args)
             state = State.from_state(state)
             if encoding == ::Encoding::UTF_8
+              unless valid_encoding?
+                raise GeneratorError, "source sequence is illegal/malformed utf-8"
```
eregon, line 454: We also need it in fast_serialize_string: https://github.com/ruby/json/pull/633

PR #609, opened 2024-10-14
Restore support for improperly encoded strings

casperisfine:
Since the `json` gem was initially written for Ruby 1.8 before strings had encoding, it used to do its own UTF-8 validation direction on bytes and never really considered the string declared encoding. So passing a ASCII-8BIT string worked as long as the bytes were valid UTF-8

We may want to deprecate this, but we should emit warnings first.

Ref: https://github.com/ruby/json/pull/595#issuecomment-2400412544

FYI: @Earlopain

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -337,20 +337,9 @@ module JSON
         # Assumes !@ascii_only, !@script_safe
         if Regexp.method_defined?(:match?)
           private def fast_serialize_string(string, buf) # :nodoc:
-            if string.encoding == ::Encoding::UTF_8
-              unless string.valid_encoding?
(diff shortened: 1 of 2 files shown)


casperisfine:
I pushed [c5a6d80](https://github.com/ruby/json/pull/609/commits/c5a6d8042752dca312cfb407a6d3e594458d1cd6), because after reflection is doesn't make sense to introduce what I think is a bug in the ruby and java versions.

Only the C version was this liberal, so only on this version we should make sure to keep supporting it for a while and probably deprecate it.

casperisfine:
Yep. That's why I think we should get rid of that behavior, but not in a minor like this. And ideally we'd warn first.

PR #610, opened 2024-10-15
Added BSDL and LEGAL files same as ruby/ruby

hsbt:
`LICENSE` file mentioned `LEGAL` and `BSDL` files. But this repository didn't have that files.

### DIFF
--- a/LEGAL
+++ b/LEGAL
@@ -1,0 +1,60 @@
+# -*- rdoc -*-
+
+= LEGAL NOTICE INFORMATION
+--------------------------
+
+All the files in this distribution are covered under either the Ruby's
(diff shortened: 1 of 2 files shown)


LEGAL
```diff
@@ -0,0 +1,60 @@
+# -*- rdoc -*-
+
+= LEGAL NOTICE INFORMATION
+--------------------------
+
+All the files in this distribution are covered under either the Ruby's
+license (see the file COPYING) or public-domain except some files
```
eregon, line 7: There is no `COPYING` file, but there is `LICENSE`

eregon, line 7: > except some files mentioned below.

The files are not mentioned, to which files the licenses below apply?

PR #614, opened 2024-10-16
Fix some typos

ydah:


benchmark/data/citm_catalog.json
```diff
@@ -1406,7 +1406,7 @@
             "description": null,
             "id": 138586671,
             "logo": "/images/UE0AAAAACEKqLwAAAAlDSVRN",
-            "name": "Boris Godounov - M.Moussorgski (version de concert)",
+            "name": "Boris Godunov - M.Moussorgski (version de concert)",
```
zverok, line 1409: As far as I understand, it was correct _French_ spelling (as well as "Moussorgski", and supported by "version de concert"), [for ex.](https://philharmoniedeparis.fr/en/activity/concert/13358-boris-godounov-orchestre-national-du-capitole-de-toulouse-tugan-sokhiev); in English that would be "[Boris Godunov - M.Mussorgsky](https://en.wikipedia.org/wiki/Boris_Godunov_(opera))"

Not that it matters anything actually :shrug:

PR #562, opened 2023-12-27
Make `JSON.generate` 1.75x as fast

mame:
This PR speeds up `JSON.generate` by approximately 1.75x (485k instructions per second -> 840k instructions per second) for [the benchmark of Oj](https://www.ohler.com/dev/oj_misc/performance_compat.html). This makes `JSON.generate` nearly as fast as `Oj.dump`.

Before:

```
$ ruby --yjit -Ilib -I ext oj-bench.rb
** Oj version 3.16.3 **
ruby 3.4.0dev (2023-12-27T05:30:20Z master 862cfcaf75) +YJIT [x86_64-linux]
Warming up --------------------------------------
             Oj.dump    83.089k i/100ms
    Oj.dump [compat]    72.410k i/100ms
     Oj.dump [rails]    57.698k i/100ms
       JSON.generate    49.706k i/100ms
Calculating -------------------------------------
             Oj.dump    836.635k (± 0.4%) i/s -     12.630M in  15.095866s
    Oj.dump [compat]    718.031k (± 0.2%) i/s -     10.789M in  15.026031s
     Oj.dump [rails]    573.882k (± 0.3%) i/s -      8.655M in  15.081085s
       JSON.generate    484.650k (± 1.0%) i/s -      7.307M in  15.078021s
```

After:

```
$ ruby --yjit -Ilib -I ext oj-bench.rb
** Oj version 3.16.3 **
ruby 3.4.0dev (2023-12-27T05:30:20Z master 862cfcaf75) +YJIT [x86_64-linux]
Warming up --------------------------------------
             Oj.dump    83.349k i/100ms
    Oj.dump [compat]    71.605k i/100ms
     Oj.dump [rails]    57.058k i/100ms
       JSON.generate    84.498k i/100ms
Calculating -------------------------------------
             Oj.dump    837.304k (± 0.3%) i/s -     12.586M in  15.031372s
    Oj.dump [compat]    718.657k (± 0.5%) i/s -     10.812M in  15.045573s
     Oj.dump [rails]    565.824k (± 0.3%) i/s -      8.502M in  15.025348s
       JSON.generate    839.614k (± 0.3%) i/s -     12.675M in  15.096044s
```

This PR consists of the following several improvements.

* Drop prebuild of `array_delim`, etc.
  * `array_delim` is usually a single comma character. Using `memcpy` to copy a single character was inefficient.
  * It is much faster to output a comma and (optional) `array_nl` separately without prebuild.
[...]

PR #620, opened 2024-10-17
Add a fast path for ASCII strings

casperisfine:
Fix: https://github.com/ruby/json/pull/562

This optimization is based on a few assumptions:

  - Most strings are ASCII only.
  - Most strings had their coderange scanned already.

If the above is true, then by checking the string coderange, we can use a much more streamlined function to encode ASCII strings.

Before:

```
ruby 3.4.0preview2 (2024-10-07 master 32c733f57b) +YJIT +PRISM [arm64-darwin23]
                json   140.000 i/100ms
                  oj   230.000 i/100ms
           rapidjson   108.000 i/100ms
                json      1.464k (± 1.4%) i/s  (682.83 μs/i) -      7.420k in   5.067573s
                  oj      2.338k (± 1.5%) i/s  (427.64 μs/i) -     11.730k in   5.017336s
           rapidjson      1.075k (± 1.6%) i/s  (930.40 μs/i) -      5.400k in   5.025469s

Comparison:
                json:     1464.5 i/s
                  oj:     2338.4 i/s - 1.60x  faster
           rapidjson:     1074.8 i/s - 1.36x  slower

```

After:

```
ruby 3.4.0preview2 (2024-10-07 master 32c733f57b) +YJIT +PRISM [arm64-darwin23]
                json   189.000 i/100ms
                  oj   228.000 i/100ms
           rapidjson   108.000 i/100ms
                json      1.903k (± 1.2%) i/s  (525.55 μs/i) -      9.639k in   5.066521s
                  oj      2.306k (± 1.3%) i/s  (433.71 μs/i) -     11.628k in   5.044096s
           rapidjson      1.069k (± 2.4%) i/s  (935.38 μs/i) -      5.400k in   5.053794s

Comparison:
                json:     1902.8 i/s
                  oj:     2305.7 i/s - 1.21x  faster
           rapidjson:     1069.1 i/s - 1.78x  slower
```

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -117,7 +117,29 @@ static void convert_UTF8_to_JSON(FBuffer *out_buffer, VALUE in_string, bool out_
     RB_GC_GUARD(in_string);
 }
 
-static void convert_ASCII_to_JSON(FBuffer *out_buffer, VALUE str, bool out_script_safe)
+static const bool escape_table[256] = {
+    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+    0,0,1,0,0,0,0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* '"'  */
+    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0,0,0,1,0,0,0, /* '\\' */
+    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
+    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1
+};
+
+static const bool script_safe_escape_table[256] = {
+    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
+    0,0,1,0,0,0,0,0,0,0,0,0,0,0,0,1, 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, /* '"' and '/' */
+    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0,0,0,1,0,0,0, /* '\\' */
+    0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0, 0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
+    1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1, 1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,1,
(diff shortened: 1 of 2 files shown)


casperisfine:
I pushed an extra commit that re-use the escape table from https://github.com/ruby/json/pull/562/commits/a81ec4770af4a2f20a9dc06d0295cf5b93a7af91

```
ruby 3.4.0preview2 (2024-10-07 master 32c733f57b) +YJIT +PRISM [arm64-darwin23]
                json   224.000 i/100ms
                  oj   230.000 i/100ms
           rapidjson   107.000 i/100ms
                json      2.254k (± 1.6%) i/s  (443.69 μs/i) -     11.424k in   5.069999s
                  oj      2.318k (± 1.4%) i/s  (431.32 μs/i) -     11.730k in   5.060421s
           rapidjson      1.081k (± 1.9%) i/s  (925.05 μs/i) -      5.457k in   5.049738s

Comparison:
                json:     2253.8 i/s
                  oj:     2318.5 i/s - same-ish: difference falls within error
           rapidjson:     1081.0 i/s - 2.08x  slower
```

For now it's only applied to the ASCII fast path, but I think we could optimize UTF8 too with the assumption that most bytes in an UTF8 string are ASCII. But I'd rather not complexify this PR more.

ext/json/ext/generator/generator.c
```diff
@@ -42,46 +41,196 @@ static void convert_UTF8_to_JSON(FBuffer *out_buffer, VALUE in_string, bool out_
...
+    for (pos = 0; pos < len;) {
+        unsigned char ch = ptr[pos];
+        /* JSON encoding */
+        if (escape_table[ch]) {
+            if (pos > beg) {
+                fbuffer_append(out_buffer, &ptr[beg], pos - beg);
+            }
+
+            beg = pos + 1;
+            switch (ch) {
```
nvasilevski, line 161: Thanks for elaborating on this change in [your blog post ](https://byroot.github.io/ruby/json/2024/12/15/optimizing-ruby-json-part-1.html) ❤️ 

I was wondering if we could entirely eliminate this `switch` at an expense of static memory by storing escaped values instead of booleans in the lookup table. Just as an example, in ruby:

```ruby
JSON_ESCAPE_TABLE['"'] = "\\\"
JSON_ESCAPE_TABLE['\\'] = "\\\\"
JSON_ESCAPE_TABLE['\b'] = "\\b"
# ...
# default: case can be pre-calculated as well for other characters
```

It may not result in a visible improvement in realistic benchmarks due to having not that many chars to escape but could be visible on a micro-benchmark specifically crafted to measure escaping performance at no cost (except static memory) for the main flow.

I could be missing something that makes it hard to store non-boolean values in `escape_table` but would love to at least learn about this as well. Thanks!

byroot: It could be worth trying, but there's a couple things making this not an obvious win.

Escaped values in this case aren't scalar. You can either store them as a table of `char *`:

```ruby
static const char *[256] escape_table = {
  0, 0, 0, "\\n", // ....
};
```

If you do this:

  - `char *` is `8B` on 64-bit platform, whereas `char` is just `1B`, so we went from `256B` to `2kiB`, not nothing.
  - `char *` doesn't store the length, so presumably to append these you'd need to call `strlen`, they are shot string so probably not very slow, but given we're just trying to eliminate a case, it may actually be slower.
  - `char *` isn't stored contiguiously, so that's another memory region to keep in CPU cache.

Alternatively, you could store the escaped strings inline, e.g.:

```c
struct escaped_char {
  unsigned char len;
  char bytes[8];
}

static const struct escaped_char[256] escape_table = {
  {0, 0}, {0, 0}, {2, "\\n"}, // ....
}
```

But this would make the escape table even bigger.

Maybe you could only do this for strings that are known to be pure-ASCII so the table only has `128` entries, keeping the memory usage down, but then you can't really re-use it for UTF-8.

Note that all of these are just "static assumption", performance isn't always easy to predict, so maybe your suggestion would actually help.

headius:
Did this get applied to the Java parser anywhere? I can take it if not but I don't want to dive in and waste my time if it's already there.

byroot:
As mentioned previously, I don't know enough about Java to do much to it, and I don't really have any incentive either.

If you want to speedup the Java version feel free to do it.

headius:
@byroot Ok thanks for clarifying. You've done so many PRs lately I lost track!

I'll do the port and push a PR. The code in the Java extension will be pretty much the same as what you wrote for the C version.

headius:
Good news: much of the structure of the Java code already mimics the C code, and a profile of the twitter.json file dumping shows most of the overhead is in two places: handling each character (because it's using heavier UTF-8 logic always) and picking the right logic for each object type (bad lookup of handler logic based on the Ruby class, most likely).

Ought to be able to speed this up nicely.

headius:
Work is proceeding on this and other optimizations for the JRuby extension in https://github.com/ruby/json/pull/725.

PR #621, opened 2024-10-17
Fix State#max_nesting=

peterzhu2118:
Returning state->max_nesting is not valid because it's not a Ruby object.


byroot (approved):
🤦 thank you!

byroot:
Holy cow, that bug has been in there for 15 years: https://github.com/ruby/json/commit/07d786882cd0a9d9131096bfcbc613f8262e1fc0

Issue #534, opened 2023-07-27
Shorten reference to input data in error message


forthrin:
The "unexpected token" error message includes the full bulk of the input data no matter how large. Suggest to snip input data at a reasonable cutoff point, eg. 1024 bytes, to prevent swamping terminals, logs, cron emails, etc.

PR #625, opened 2024-10-18
Limit the size of ParserError exception messages

casperisfine:
Fix: https://github.com/ruby/json/issues/534

Only include up to 32 bytes of the unparseable source.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,7 @@
 # Changes
 
 * Numerous performance optimizations in `JSON.generate` and `JSON.dump`.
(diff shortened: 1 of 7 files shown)


headius:
Thank you for continuing to patch the JRuby ext as well! Please ping me if there's any bug or optimization you think I should look into.

casperisfine:
Note that I'm only doing the minimum required to keep the test suite passing, I'm not looking at all at the performance of it.

On that front I consider you to be the maintainer of the Java backend.

casperisfine:
Not that I don't want to, but my Java skills are primitive at best and it's not in my plan to spend the time to develop a Java expertise.

PR #629, opened 2024-10-18
Optimize convert_UTF8_to_JSON for mostly ASCII strings

casperisfine:
If we assume that even UTF-8 strings are mostly ASCII, we can implement a fast path for the ASCII parts.

Before:

```
== Encoding mixed utf8 (20012001 bytes)
ruby 3.4.0dev (2024-10-18T15:12:54Z master d1b5c10957) +YJIT +PRISM [arm64-darwin23]
                json     5.000 i/100ms
                  oj     9.000 i/100ms
           rapidjson     2.000 i/100ms
                json     49.403 (± 2.0%) i/s   (20.24 ms/i) -    250.000 in   5.062647s
                  oj    100.120 (± 2.0%) i/s    (9.99 ms/i) -    504.000 in   5.035349s
           rapidjson     26.404 (± 0.0%) i/s   (37.87 ms/i) -    132.000 in   5.001025s

Comparison:
                json:       49.4 i/s
                  oj:      100.1 i/s - 2.03x  faster
           rapidjson:       26.4 i/s - 1.87x  slower
```

After:

```
== Encoding mixed utf8 (20012001 bytes)
ruby 3.4.0dev (2024-10-18T15:12:54Z master d1b5c10957) +YJIT +PRISM [arm64-darwin23]
                json    10.000 i/100ms
                  oj     9.000 i/100ms
           rapidjson     2.000 i/100ms
                json     95.686 (± 2.1%) i/s   (10.45 ms/i) -    480.000 in   5.018575s
                  oj     96.875 (± 2.1%) i/s   (10.32 ms/i) -    486.000 in   5.019097s
           rapidjson     26.260 (± 3.8%) i/s   (38.08 ms/i) -    132.000 in   5.033151s

Comparison:
                json:       95.7 i/s
                  oj:       96.9 i/s - same-ish: difference falls within error
           rapidjson:       26.3 i/s - 3.64x  slower
```

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -71,7 +71,7 @@ static void convert_UTF8_to_JSON(FBuffer *out_buffer, VALUE str, const char esca
                 }
                 case 3: {
                     unsigned char b2 = ptr[pos + 1];
-                    if (out_script_safe && b2 == 0x80) {
+                    if (RB_UNLIKELY(out_script_safe && b2 == 0x80)) {
                         unsigned char b3 = ptr[pos + 2];
                         if (b3 == 0xA8) {
                             FLUSH_POS(3);
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -25,7 +25,7 @@ static ID i_to_s, i_to_json, i_new, i_pack, i_unpack, i_create_id, i_extend;
  * Everything else (should be UTF-8) is just passed through and
  * appended to the result.
  */
-static void convert_UTF8_to_JSON(FBuffer *out_buffer, VALUE str, const bool escape_table[256], bool out_script_safe)
+static void convert_UTF8_to_JSON(FBuffer *out_buffer, VALUE str, const char escape_table[256], bool out_script_safe)
 {
     const char *hexdig = "0123456789abcdef";
     char scratch[12] = { '\\', 'u', 0, 0, 0, 0, '\\', 'u' };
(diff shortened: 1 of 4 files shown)


ext/json/ext/generator/generator.c
```diff
@@ -25,96 +25,77 @@ static ID i_to_s, i_to_json, i_new, i_pack, i_unpack, i_create_id, i_extend;
...
+                case '\t': FLUSH_POS(1); fbuffer_append(out_buffer, "\\t", 2); break;
+                default: {
+                    if ((ch & 0x80) == 0x00) { /* leading 1 bit is   0b0     */
+                        FLUSH_POS(1);
                         scratch[2] = hexdig[ch >> 12];
                         scratch[3] = hexdig[(ch >> 8) & 0xf];
                         scratch[4] = hexdig[(ch >> 4) & 0xf];
                         scratch[5] = hexdig[ch & 0xf];
                         fbuffer_append(out_buffer, scratch, 6);
+                    } else if ((ch & 0xE0) == 0xC0) { /* leading 3 bits are 0b110   */
```
casperisfine, line 60: Makes me wonder if we could actually store the character length in the escape table, and directly jump. 🤔

byroot: 31d811faf1baafb75c7bb5de0b20bae28a277c63

eregon: One thing to care is this doesn't guarantee there are 1-6 bytes left in the String if the String is broken.

byroot: We never enter this path if the string is broken: https://github.com/ruby/json/blob/9ce215d5e1e4a07d67036238a33924323fa78383/ext/json/ext/generator/generator.c#L739

We rely on the string coderange to guarantee us we don't have to deal with broken UTF-8.

The assumption is that most strings have their coderange known. If we wanted to avoid scanning the coderange when it is unknown we could have an alternative function that does protect against broken strings, but not sure if it's worth it.

eregon: Ah right a broken String is just an exception on JSON.dump (unlike on JSON.load), nice.

PR #633, opened 2024-10-21
Raise the correct exception in fast_serialize_string

eregon:
Fix: https://github.com/ruby/json/issues/634

* Related to https://github.com/ruby/json/issues/344

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -339,6 +339,7 @@ module JSON
           private def fast_serialize_string(string, buf) # :nodoc:
             buf << '"'
             string = string.encode(::Encoding::UTF_8) unless string.encoding == ::Encoding::UTF_8
+            raise GeneratorError, "source sequence is illegal/malformed utf-8" unless string.valid_encoding?
 
             if /["\\\x0-\x1f]/n.match?(string)
               buf << string.gsub(/["\\\x0-\x1f]/n, MAP)
--- a/test/json/json_generator_test.rb
+++ b/test/json/json_generator_test.rb
@@ -449,16 +449,12 @@ EOT
     end
     assert_includes error.message, "source sequence is illegal/malformed utf-8"
 
-    # These pass on the pure-Ruby generator but not with the native extension
-    # https://github.com/ruby/json/issues/634
-    if defined?(JSON::Pure)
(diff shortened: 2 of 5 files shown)


lib/json/pure/generator.rb
```diff
@@ -339,6 +339,7 @@ def generate(obj)
           private def fast_serialize_string(string, buf) # :nodoc:
             buf << '"'
             string = string.encode(::Encoding::UTF_8) unless string.encoding == ::Encoding::UTF_8
+            raise GeneratorError, "source sequence is illegal/malformed utf-8" unless string.valid_encoding?
```
byroot, line 342: Probably need to do it in the Ruby 2.3 version too?

test/json/json_generator_test.rb
```diff
@@ -443,6 +443,20 @@ def test_invalid_encoding_string
...
+    assert_includes error.message, "source sequence is illegal/malformed utf-8"
+
+    # These pass on the pure-Ruby generator but not with the native extension
+    # assert_raise(Encoding::UndefinedConversionError) do
+    #   "\x82\xAC\xEF".b.to_json
+    # end
+    #
+    # assert_raise(Encoding::UndefinedConversionError) do
+    #   JSON.dump("\x82\xAC\xEF".b)
+    # end
```
byroot, line 459: You can conditionally run it with `if defined?(JSON::Pure)`.

I'll be trying to make implementations converge over time, but short term it's helpful to test the divergent behavior.

eregon, line 459: I filed https://github.com/ruby/json/issues/634 for this

lib/json/pure/generator.rb
eregon: That one calls `String#to_json` so it's fine

test/json/json_generator_test.rb
eregon: OK, done.

byroot: Actually there is already a test for it, so just remove the commented out code I think: https://github.com/ruby/json/blob/f887daa403a7f5ad880c48ef3f6c72615df08a80/test/json/json_generator_test.rb#L448-L456

eregon: That's different, it's for valid UTF-8 bytes, this is invalid UTF-8 bytes.

PR #637, opened 2024-10-24
Use `String#b` and `String#+`

nobu:


byroot:
Oh, for some reason I thought Ruby 2.3 didn't have `String#+@`

PR #639, opened 2024-10-24
Reduce CI

### LINKED ISSUE #638: Reduce CI
Currently there are 72 workflows except for ruby-versions.
This seems too many, and old platforms don't feel to be worth to test, such as macos-12 and ubuntu-20.04.
Also is it necessary to keep tests for old ruby versions?

nobu:
Fix #638.
- Drop old platforms, ubuntu-20.04 and macos-12.
- Drop old ruby versions, 2.4..2.6.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,9 +20,7 @@ jobs:
       fail-fast: false
       matrix:
         os:
-        - ubuntu-20.04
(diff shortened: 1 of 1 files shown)


byroot:
Yes, this was on my radar. I'll still release 2.7.3 with current support, but yes, dropping older rubies make sense to me for 2.8.0.

Issue #642, opened 2024-10-24
Emit a deprecation warning when dealing with incorrectly encoded strings


byroot:
Ref: https://github.com/ruby/json/issues/138

Because it was initially implemented for Ruby 1.8 which didn't have proper string encoding, in a bunch of places the C version does silently accept strings with broken encodings.

We should emit a deprecation warning in 2.8.0, and remove that mis-feature in 3.0

  - [ ] https://github.com/ruby/json/blob/fb25e94aeab510ec67be44ab61ae664aa67e62dd/ext/json/ext/parser/parser.rl#L693
  - [ ] https://github.com/ruby/json/blob/fb25e94aeab510ec67be44ab61ae664aa67e62dd/ext/json/ext/generator/generator.c#L768

Issue #138, opened 2012-05-31
C and Pure parser accept invalid UTF-8 strings, the Java parser doesn't.


headius:
I know, I know...if it's bad content it's bad content. But this represents a difference from MRI.

Here's the case, again a reduced version of one I got from @rkh:

``` ruby
# encoding: utf-8
require 'json'
x = "{\"foo\":\"\xC3\"}"
h = JSON.parse(x)
p h['foo']
p h['foo'].encoding
```

So basically there's a bad byte in a UTF-8 string, and the MRI version walks right by it and allows it to come through to the resulting parsed json structure.

I have a totally broken patch for this:

```
diff --git a/java/src/json/ext/ByteListTranscoder.java b/java/src/json/ext/ByteListTranscoder.java
index ed9e54b..a7e42ba 100644
--- a/java/src/json/ext/ByteListTranscoder.java
+++ b/java/src/json/ext/ByteListTranscoder.java
@@ -78,9 +78,10 @@ abstract class ByteListTranscoder {
             return head;
         }
         if (head <= 0xbf) { // 0b10xxxxxx
-            throw invalidUtf8(); // tail byte with no head
+            return head; //throw invalidUtf8(); // tail byte with no head
         }
         if (head <= 0xdf) { // 0b110xxxxx
+            if (pos + 1 > srcEnd) return head;
             ensureMin(1);
             int cp = ((head  & 0x1f) << 6)
                      | nextPart();
```

Again, I'm not sure this is actually something that needs to be fixed, but because the MRI version of json does _not_ blow up on this content, there's something to be addressed.

headius:
It would not be possible for us to implement such a config flag because the parser we use is always strict about this. @flori said above they implemented the error for 1.9+ pure and ext, so is there really anything to do here 12 years later?

byroot:
As far as I can tell, it's still not raising today:

```ruby
>> JSON::Pure::Parser.new("{\"foo\":\"\xC3\"}").parse["foo"].encoding
=> #<Encoding:UTF-8>
>> JSON::Pure::Parser.new("{\"foo\":\"\xC3\"}").parse["foo"].valid_encoding?
=> false
>> JSON::Ext::Parser.new("{\"foo\":\"\xC3\"}").parse["foo"].encoding
=> #<Encoding:UTF-8>
>> JSON::Ext::Parser.new("{\"foo\":\"\xC3\"}").parse["foo"].valid_encoding?
=> false
```

headius:
[...]

PR #643, opened 2024-10-24
Emit warnings when dumping binary strings

casperisfine:
Fix: https://github.com/ruby/json/issues/642

Because of it's Ruby 1.8 heritage, the C extension doesn't care much about strings encoding. We should get stricter over time.

### DIFF
--- a/ext/json/ext/parser/parser.rl
+++ b/ext/json/ext/parser/parser.rl
@@ -693,9 +693,12 @@ static VALUE convert_encoding(VALUE source)
 
  if (encindex == binary_encindex) {
     // For historical reason, we silently reinterpret binary strings as UTF-8 if it would work.
-    // TODO: Deprecate in 2.8.0
-    // TODO: Remove in 3.0.0
-    return rb_enc_associate_index(rb_str_dup(source), utf8_encindex);
+    VALUE utf8_string = rb_enc_associate_index(rb_str_dup(source), utf8_encindex);
+    switch (rb_enc_str_coderange(utf8_string)) {
+      case ENC_CODERANGE_7BIT:
+      case ENC_CODERANGE_VALID:
+        return utf8_string;
+    }
   }
 
   return rb_str_conv_enc(source, rb_enc_from_index(encindex), rb_utf8_encoding());
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -46,6 +46,8 @@ static void generate_json_fixnum(FBuffer *buffer, struct generate_json_data *dat
 static void generate_json_bignum(FBuffer *buffer, struct generate_json_data *data, JSON_Generator_State *state, VALUE obj);
 static void generate_json_float(FBuffer *buffer, struct generate_json_data *data, JSON_Generator_State *state, VALUE obj);
 
+static int usascii_encindex, utf8_encindex, binary_encindex;
+
 /* Converts in_string to a JSON string (without the wrapping '"'
  * characters) in FBuffer out_buffer.
  *
@@ -535,7 +537,7 @@ static VALUE mString_to_json_raw_object(VALUE self)
     VALUE result = rb_hash_new();
     rb_hash_aset(result, rb_funcall(mJSON, i_create_id, 0), rb_class_name(rb_obj_class(self)));
     ary = rb_funcall(self, i_unpack, 1, rb_str_new2("C*"));
-    rb_hash_aset(result, rb_str_new2("raw"), ary);
+    rb_hash_aset(result, rb_utf8_str_new_lit("raw"), ary);
     return result;
 }
 
@@ -822,8 +824,6 @@ static void generate_json_array(FBuffer *buffer, struct generate_json_data *data
     fbuffer_append_char(buffer, ']');
 }
 
-static int usascii_encindex, utf8_encindex, binary_encindex;
-
 static inline int enc_utf8_compatible_p(int enc_idx)
 {
     if (enc_idx == usascii_encindex) return 1;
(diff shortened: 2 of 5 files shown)


ext/json/ext/parser/parser.rl
```diff
@@ -690,9 +690,12 @@ static VALUE convert_encoding(VALUE source)
...
     // For historical reason, we silently reinterpret binary strings as UTF-8 if it would work.
-    // TODO: Deprecate in 2.8.0
-    // TODO: Remove in 3.0.0
-    return rb_enc_associate_index(rb_str_dup(source), utf8_encindex);
+    VALUE utf8_string = rb_enc_associate_index(rb_str_dup(source), utf8_encindex);
+    switch (rb_enc_str_coderange(utf8_string)) {
+      case ENC_CODERANGE_7BIT:
+      case ENC_CODERANGE_VALID:
+        return utf8_string;
+    }
```
casperisfine, line 698: This one is https://github.com/ruby/json/issues/138. I'm having second thoughts about deprecating / removing it, as I fear lots of people depend on it, and looking at other parsers they all seem to accept it. So maybe we should just support it in the Java parser? 

WDYT @headius @eregon ?

headius: Whoops, I commented too early and thought this was a YAML issue. 

After reexamining the code, it does appear that the json implementation for JRuby works directly with bytes via the ragel parser. It would probably be possible to make it accept invalid characters. 

My position remains the same, however. I'm not sure what other person is you tested, but all json specifications I looked at explicitly require content to be in one of the main unicode UTF encodings. That says to me that providing invalid characters is off spec and should produce an error.

The original RFC from 2006 even says:

> JSON text SHALL be encoded in Unicode. The default encoding is UTF-8.

https://www.ietf.org/rfc/rfc4627.txt

The character stream cannot be parsed as Unicode, then it is not up to spec.

byroot: > I'm not sure what other parsers you tested,

`oj` and Python's stdlib.

I totally get your argument, and generally agree, the concern is really about breaking existing (misbehaving) code.

headius: Maybe I'm crazy, but I would break it in a minor release so it's easy to revert with another minor release, rather than break it in a major release and have to go back. I doubt the impact will be large. If we put out a minor release that starts erroring on bad input and it doesn't get reported for a few months, it's probably fine. 

I think it's also important to remember that Ruby has become much more strict about character encodings over the years so the chance of getting bad content has reduced significantly since the original issue.

casperisfine: A common case where this still happens a lot today is JSON POST requests handled by Rack.

The Rack spec clearly state the body should be `ASCII_8BIT`, so if you do the common `JSON.parse(request.body)` you hit this case.

That's why I'm worried about breaking this.

byroot:
Merging as is for now, we can always decide to be even more string in a followup.

ext/json/ext/parser/parser.rl
eregon: Isn't #138 another case, specifically it seems about a broken UTF-8 string passed to `JSON.parse`, while the code here seems to be about a BINARY string which would be invalid if forced to UTF-8.

I think it would be best to treat BINARY forced to UTF-8 as orthogonal to whether broken UTF-8 bytes are allowed or not in JSON.parse. IOW I think we shouldn't add this check here (unless it's already done for broken UTF-8 strings too).
I think the rb_str_conv_enc() below doesn't check for broken characters, does it?

IMO a broken UTF-8 string is still a UTF-8 string, and Ruby generally allows broken strings in many places so unless there is a compelling reason to change that I would just allow them. in the end we are copying bytes + interpreting escape sequences, the string being broken doesn't prevent us to do that and I don't think it makes it much slower either.

byroot: Yeah that's a decent point. 

Just yesterday someone at work told me they had an outage because they have a JSON based logger and were logging binary strings (some avro stuff or whatever).

So perhaps you are right and should just keep this behavior, unless `strict: true` is set.

eregon: > unless `strict: true` is set.

https://github.com/ruby/json/blob/6d3b3ac9b4ecceebc230da73d530c45d231b763f/lib/json/pure/generator.rb#L227-L228

I think even strict: true should accept it given the documented meaning of it.
It could be another option, but not sure there is any value to prohibit broken strings if anyway it works fine.

byroot: Would you be willing to make a PR with a test of what you have in mind? I can make it pass, just want to be 100% sure we're talking about the same thing.

eregon: What I'm thinking is:
```diff
diff --git a/test/json/json_parser_test.rb b/test/json/json_parser_test.rb
index 59cfcfa..a7ec04d 100644
--- a/test/json/json_parser_test.rb
+++ b/test/json/json_parser_test.rb
@@ -196,6 +196,19 @@ class JSONParserTest < Test::Unit::TestCase
     )
   end
 
+  def test_parse_broken_string
+    # https://github.com/ruby/json/issues/138
+    s = parse(%{["\x80"]})[0]
+    assert_equal("\x80", s)
+    assert_equal Encoding::UTF_8, s.encoding
+    assert_equal false, s.valid_encoding?
+
+    s = parse(%{["\x80"]}.b)[0]
+    assert_equal("\x80", s)
+    assert_equal Encoding::UTF_8, s.encoding
+    assert_equal false, s.valid_encoding?
+  end
+
   def test_parse_big_integers
     json1 = JSON(orig = (1 << 31) - 1)
     assert_equal orig, parse(json1)
```

And that passes on master, because of the `rb_str_conv_enc()` below which if there is an error during conversion returns the source string as-is. And I guess the C extension doesn't really care whether the input String is internally in UTF-8 or BINARY and just works on bytes (although that could lead to subtle bugs as e.g. `ENC_CODERANGE_VALID` means something different for both encodings).
If I replace the `rb_str_conv_enc()` with `return rb_funcall(source, rb_intern("encode"), 1, rb_const_get(rb_path2class("Encoding"), rb_intern("UTF_8")));` then I get `Encoding::UndefinedConversionError`.
If on top of that I undo the change in this file then it works again.

eregon: https://github.com/ruby/json/pull/665

Issue #647, opened 2024-10-25
Installing to vendor/bundle appears to be broken


pixeltrix:
Whilst this appears to be similar to #646 this app does not have `json_pure` in the bundle.

Running `bundle install` inside a Debian Bullseye based docker image works fine, but running `bundle install --path vendor/bundle` breaks with the following error:

```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

current directory:
/app/vendor/bundle/ruby/3.0.0/gems/appsignal-3.13.1/ext
/usr/local/bin/ruby -I /usr/local/lib/ruby/3.0.0 -r
./siteconf20241025-15-5udun9.rb extconf.rb
/usr/local/lib/ruby/3.0.0/json/common.rb:82:in `initialize': wrong number
of arguments (given 1, expected 0) (ArgumentError)
	from /usr/local/lib/ruby/3.0.0/json/common.rb:82:in `new'
	from /usr/local/lib/ruby/3.0.0/json/common.rb:82:in `create_fast_state'
	from /usr/local/lib/ruby/3.0.0/json/common.rb:75:in `generator='
	from /usr/local/lib/ruby/3.0.0/json/ext.rb:11:in `<module:Ext>'
	from /usr/local/lib/ruby/3.0.0/json/ext.rb:6:in `<module:JSON>'
	from /usr/local/lib/ruby/3.0.0/json/ext.rb:3:in `<top (required)>'
from
<internal:/usr/local/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in
`require'
from
<internal:/usr/local/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in
`require'
	from /usr/local/lib/ruby/3.0.0/json.rb:579:in `<module:JSON>'
	from /usr/local/lib/ruby/3.0.0/json.rb:575:in `<top (required)>'
from
<internal:/usr/local/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in
`require'
from
<internal:/usr/local/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in
`require'
from /app/vendor/bundle/ruby/3.0.0/gems/appsignal-3.13.1/ext/base.rb:7:in
`<top (required)>'
from
<internal:/usr/local/lib/ruby/3.0.0/rubygems/core_ext/kernel_require.rb>:85:in
`require'
from
[...]

Issue #646, opened 2024-10-24
json 2.7.3 has broken my test environment


david22swan:
Hey, sorry to bother but have had issues with the newest release of the Gem with the below error blocking my test development:

```
An error occurred while loading ./spec/acceptance/set_config_spec.rb.
On host `localhost:52362'
Failure/Error: require 'json'
ArgumentError:
  wrong number of arguments (given 1, expected 0)

# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json/common.rb:82:in `initialize'
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json/common.rb:82:in `new'
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json/common.rb:82:in `create_fast_state'
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json/common.rb:75:in `generator='
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json/ext.rb:11:in `<module:Ext>'
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json/ext.rb:6:in `<module:JSON>'
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json/ext.rb:3:in `<top (required)>'
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json.rb:579:in `<module:JSON>'
# ./.bundle/ruby/3.2.0/gems/json_pure-2.6.3/lib/json.rb:575:in `<top (required)>'
# ./spec/spec_helper_acceptance.rb:8:in `<top (required)>'
# ./spec/acceptance/set_config_spec.rb:1:in `<top (required)>'
```

Was able to resolve this by pinning the json gem itself to 2.7.2 so can confirm the issue is with the new release, and thought I should shout out incase anyone else has gotten hit with it.
Unsure of the actual cause of the error, or if it has something to do with us using json_pure and json being pulled in from another dependency.

byroot:
Ah yes, `json_pure` and `json` need to be at the same version if both are co-existing.

david22swan:
Was able to fix the issue with them different though.
Fixed it by pinning json to 2.7.2 while json_pure was still pinned to 2.6.3?

byroot:
Yes, some version mix can sometime work, but they are only guaranteed to work if the version matches.

[...]

PR #648, opened 2024-10-25
Workaround rubygems $LOAD_PATH bug

casperisfine:
Ref: https://github.com/ruby/json/issues/647
Ref: https://github.com/rubygems/rubygems/pull/6490

Older rubygems are executing `extconf.rb` with a broken `$LOAD_PATH` causing the `json` gem native extension to be loaded with the stdlib version of the `.rb` files.

This fails with

```
json/common.rb:82:in `initialize': wrong number of arguments (given 1, expected 0) (ArgumentError)
```

Since this is just for `extconf.rb` we can probably just accept that extra argument and ignore it.

The bug was fixed in rubygems 3.4.9 / 2023-03-20

FYI: @pixeltrix

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # Changes
 
+* Workaround a bug in 3.4.8 and older https://github.com/rubygems/rubygems/pull/6490.
+* Workaround different versions of `json` and `json_pure` being loaded (not officially supported).
+* Make `json_pure` Ractor compatible. 
+
 ### 2024-10-24 (2.7.3)
 
 * Numerous performance optimizations in `JSON.generate` and `JSON.dump` (up to 2 times faster).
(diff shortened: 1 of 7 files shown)


casperisfine:
The problem is that this makes https://github.com/ruby/json/issues/646 somewhat worse, as it changes from a boot time issue into a runtime one.

```
$ bundle exec ruby -W -e 'require "json"; JSON.dump({})'
/Users/byroot/.gem/ruby/3.3.4/gems/json_pure-2.6.3/lib/json/common.rb:300: warning: The json gem extension was loaded with the stdlib ruby code. You should upgrade rubygems with `gem update --system`
/Users/byroot/.gem/ruby/3.3.4/gems/json_pure-2.6.3/lib/json/common.rb:310:in `generate': undefined method `configure' for an instance of JSON::Ext::Generator::State (NoMethodError)

      state = state.configure(opts)
                   ^^^^^^^^^^
	from /Users/byroot/.gem/ruby/3.3.4/gems/json_pure-2.6.3/lib/json/common.rb:641:in `dump'
	from -e:1:in `<main>'
```

The same code being shipped in `json` and `json_pure` is really a problem, I think we might need to namespace them somehow? Or find a way to enforce that both versions match.

PR #664, opened 2024-10-30
Skip calling configure if there are no options

eregon:
* Set all defaults in the constructor.
* Order the instance variables in the same order as in #configure.

Unfortunately because of `JSON.dump_default_options` this fast path won't apply for `JSON.dump`.
I wish those values (`max_nesting: false, allow_nan: true`) would be the default for all generate/dump-like methods and then they could be default in the State constructor.

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -142,10 +142,12 @@ module JSON
           @array_nl              = ''
           @allow_nan             = false
           @ascii_only            = false
-          @script_safe          = false
-          @strict                = false
+          @depth                 = 0
           @buffer_initial_length = 1024
-          configure(opts || {})
+          @script_safe           = false
+          @strict                = false
+          @max_nesting           = 100
+          configure(opts) if opts
         end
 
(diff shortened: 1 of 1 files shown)


eregon:
Maybe we could store the default state for `JSON.dump` in a constant and then `.dup` it, to avoid calling `configure` which seems particularly slow/heavy.
Kind of similar to the current `const_set :FAST_STATE_PROTOTYPE, create_fast_state`.
However that wouldn't respect mutations of `JSON.dump_default_options` or reassignments (`JSON.dump_default_options = { ... }`).
If we don't support direct mutation and freeze the Hash then we could still compare `JSON.dump_default_options` by identity and cache.

Though obviously any global state seems bad and maybe the best is to deprecate `JSON.dump_default_options` entirely.

byroot:
Yeah, I toyed with this idea too last week, and `JSON.dump_default_options` is indeed the blocker.

I did however managed to avoid allocating the State object entirely in https://github.com/ruby/json/pull/662 so it's less of a problem now.

Issue #634, opened 2024-10-21
JSON.dump("\x82\xAC\xEF".b) no error with the C extension


eregon:
On json 2.7.2:
```
$ ruby -rjson/ext -e 'p JSON.dump("\x82\xAC\xEF".b)'
json-2.7.2/lib/json/common.rb:306:in `generate': source sequence is illegal/malformed utf-8 (JSON::GeneratorError)
$ ruby -rjson/pure -e 'p JSON.dump("\x82\xAC\xEF".b)'
json-2.7.2/lib/json/pure/generator.rb:453:in `encode': "\x82" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
```

On json master:
```
$ ruby -rjson/ext -e 'p JSON.dump("\x82\xAC\xEF".b)'
"\"\\u0082\\u00ac\xEF\""
$ ruby -rjson/pure -e 'p JSON.dump("\x82\xAC\xEF".b)'
json-2.7.2/lib/json/pure/generator.rb:341:in `encode': "\x82" from ASCII-8BIT to UTF-8 (Encoding::UndefinedConversionError)
```

byroot:
It is known, and I don't plan to change it until 3.0.

Ref: https://github.com/ruby/json/pull/609

eregon:
I think it's due to using `obj = rb_str_export_to_enc(obj, rb_utf8_encoding());` vs using `String#encode`, so a potential fix would be to use `String#encode` to get a `Encoding::UndefinedConversionError`.
It could also be a JSON::GeneratorError like the latest release (although inconsistent with the pure-Ruby version), that would still be better than silently accepting broken strings.

byroot:
Hum, wait. I think I'm misreading your report.

byroot:
yeah, this is a regression from 2.7.2, good catch I'll fix it.

eregon:
I think it's related but not the same, https://github.com/ruby/json/pull/609 is about bytes valid in UTF-8 and this is about bytes invalid in UTF-8.

I'm not sure about the semantics of `rb_str_export_to_enc()` is it like String#encode (from `ruby -rjson/ext -e 'p JSON.dump("A".encode("UTF-32LE"))'`) but without exceptions?
What does it do if given a BINARY string with valid UTF-8 bytes?
It seems to be like `force_encoding` in that case, but unsure of the details:
`ruby -rjson/ext -e 'p JSON.dump("Aé".b)'` -> `"\"Aé\""

[...]

PR #665, opened 2024-10-30
Add test for parsing broken strings and use String#encode instead of rb_str_conv_enc() in parser

eregon:
From https://github.com/ruby/json/pull/643/files#r1822740241

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -89,14 +89,14 @@ static void raise_parse_error(const char *format, const char *start)
     rb_enc_raise(rb_utf8_encoding(), rb_path2class("JSON::ParserError"), format, ptr);
 }
 
-static VALUE mJSON, mExt, cParser, eNestingError;
+static VALUE mJSON, mExt, cParser, eNestingError, Encoding_UTF_8;
 static VALUE CNaN, CInfinity, CMinusInfinity;
 
 static ID i_json_creatable_p, i_json_create, i_create_id, i_create_additions,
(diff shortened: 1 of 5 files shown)


eregon:
This is only for the parsing part.
I think there it's fair to treat BINARY as UTF-8 unconditionally, because in the end we're just parsing bytes.
If the encoding of the source string is unknown (== BINARY) it seems rather fair to assume it's actually UTF-8 for JSON.

For the generating part I think we could do the same: `str.encoding == BINARY ? str.dup.force_encoding(UTF-8) : str.encode(UTF-8)`.
But it's less clear if that's good or bad, because we're potentially trying to dump something which cannot be dumped faithfully.
And it seems to already be an error in all previous JSON versions (except JSON 1.7.7 in Ruby 2.0!) when the String is broken, for `ruby -rjson -e 'p JSON.dump(["\x80"])'`. Let's keep that error (ref: https://github.com/ruby/json/issues/634).
When it's binary but not broken, `ruby -rjson -e 'p JSON::VERSION; p JSON.dump(["é".b])'`, it seems only since 2.6.1, 2.5.1 and before gave Encoding::UndefinedConversionError.
Only the C extension as this behavior, so I think it's good to warn there (ref: https://github.com/ruby/json/pull/609).

TLDR: I think this change is good for the parsing part, for the generating part it seems good as-is and to add the warning in https://github.com/ruby/json/issues/642.

eregon:
It's kind of funky that JSON used rb_str_export_to_enc() in #634 and rb_str_conv_enc() and both of these have error-prone semantics in case the conversion fails. At least now it should be consistent and just use String#encode, like the pure-Ruby version.

PR #666, opened 2024-10-30
update json version to correct version

ajayk:
The recently released version has a wrong version 
https://github.com/ruby/json/blob/v2.7.5/lib/json/version.rb


ajayk:
@byroot : ^^

PR #668, opened 2024-10-31
JSON.generate: call to_json on String subclasses

casperisfine:
Fix: https://github.com/ruby/json/issues/667

This is yet another behavior on which the various implementations differed, but the C implementation used to call `to_json` on String subclasses used as keys.

This was optimized out in e125072130229e54a651f7b11d7d5a782ae7fb65 but there is an Active Support test case for it, so it's best to make all 3 implementation respect this behavior.

FYI: @mtasaka

Issue #667, opened 2024-10-31
active_support test began to fail beginning with json 2.7.3


mtasaka:
This test:
https://github.com/rails/rails/blob/d0dcb8fa6073a0c4d42600c15e82e3bb386b27d3/activesupport/test/json/encoding_test.rb#L55

began to fail. git bisect shows the above began to fail with:
https://github.com/ruby/json/commit/e125072130229e54a651f7b11d7d5a782ae7fb65

Extracted self-contained test case:

```ruby
#!/usr/bin/ruby

require "minitest"
require "json"
require 'forwardable'
require 'minitest/autorun'

module ActiveSupport
  class TestCase < ::Minitest::Test
    Assertion = Minitest::Assertion

    class << self
    end

  end
end

class String
  def as_json(options = nil) # :nodoc:
    self
  end
end

class Hash
  def as_json(options = nil) # :nodoc:
    # create a subset of the hash by applying :only or :except
    subset = if options
      if attrs = options[:only]
        slice(*Array(attrs))
      elsif attrs = options[:except]
        except(*Array(attrs))
      else
        self
      end
    else
      self
    end

    result = {}
    subset.each do |k, v|
      result[k.to_s] = options ? v.as_json(options.dup) : v.as_json
    end
    result
  end
end

module ActiveSupport
  class << self
    extend Forwardable
    def_delegators :'ActiveSupport::JSON::Encoding',
      :use_standard_json_time_format, :use_standard_json_time_format=,
      :time_precision, :time_precision=,
      :escape_html_entities_in_json, :escape_html_entities_in_json=,
      :json_encoder, :json_encoder=
  end

  module JSON
    # Dumps objects in JSON (JavaScript Object Notation).
    # See http://www.json.org for more info.
    #
    #   ActiveSupport::JSON.encode({ team: 'rails', players: '36' })
    #   # => "{\"team\":\"rails\",\"players\":\"36\"}"
    def self.encode(value, options = nil)
      Encoding.json_encoder.new(options).encode(value)
    end

    module Encoding # :nodoc:
      class JSONGemEncoder # :nodoc:
        attr_reader :options

        def initialize(options = nil)
          @options = options || {}
        end

        # Encode the given object into a JSON string
        def encode(value)
          stringify jsonify value.as_json(options.dup)
        end

        private
          # Rails does more escaping than the JSON gem natively does (we
          # escape \u2028 and \u2029 and optionally >, <, & to work around
          # certain browser problems).
[...]

PR #674, opened 2024-11-01
Speedup #generate_json by using case/when/end

eregon:
* if/elsif comparing `obj.class` is significantly slower: https://github.com/ruby/json/pull/668#issuecomment-2450747190
* The only case where an exact class check is needed so far is for String (https://github.com/ruby/json/issues/667).
* Before:
```
$ ruby -Ilib:ext benchmark/standalone.rb dump pure JSON::Pure::Generator
truffleruby 24.2.0-dev-07b978e4, like ruby 3.2.4, Oracle GraalVM JVM [x86_64-linux] Calculating -------------------------------------
      JSON.dump(obj)      6.426k (± 5.9%) i/s  (155.62 μs/i) -     32.395k in   5.064479s
      JSON.dump(obj)      6.380k (± 7.4%) i/s  (156.73 μs/i) -     31.806k in   5.021304s
      JSON.dump(obj)      6.276k (±10.5%) i/s  (159.33 μs/i) -     31.217k in   5.060762s
      JSON.dump(obj)      6.450k (± 7.0%) i/s  (155.05 μs/i) -     32.395k in   5.059538s
      JSON.dump(obj)      6.413k (± 6.2%) i/s  (155.93 μs/i) -     32.395k in   5.081573s
```
* After:
```
$ ruby -Ilib:ext benchmark/standalone.rb dump pure
JSON::Pure::Generator
truffleruby 24.2.0-dev-07b978e4, like ruby 3.2.4, Oracle GraalVM JVM [x86_64-linux]
      JSON.dump(obj)      8.237k (± 5.0%) i/s  (121.41 μs/i) -     41.600k in   5.069507s
      JSON.dump(obj)      8.179k (± 5.1%) i/s  (122.26 μs/i) -     40.768k in   5.002035s
      JSON.dump(obj)      8.147k (± 7.9%) i/s  (122.74 μs/i) -     40.768k in   5.044840s
      JSON.dump(obj)      8.137k (± 6.9%) i/s  (122.90 μs/i) -     40.768k in   5.048690s
      JSON.dump(obj)      8.112k (±10.2%) i/s  (123.27 μs/i) -     39.936k in   5.023502s
```

### DIFF
--- a/lib/json/pure/generator.rb
+++ b/lib/json/pure/generator.rb
@@ -307,20 +307,18 @@ module JSON
 
         # Handles @allow_nan, @buffer_initial_length, other ivars must be the default value (see above)
         private def generate_json(obj, buf)
-          klass = obj.class
-          if klass == Hash
+          case obj
+          when Hash
             buf << '{'
             first = true
             obj.each_pair do |k,v|
               buf << ',' unless first
 
               key_str = k.to_s
-              if key_str.is_a?(::String)
-                if key_str.class == ::String
-                  fast_serialize_string(key_str, buf)
-                else
-                  generate_json(key_str, buf)
-                end
+              if key_str.class == String
+                fast_serialize_string(key_str, buf)
+              elsif key_str.is_a?(String)
+                generate_json(key_str, buf)
               else
                 raise TypeError, "#{k.class}#to_s returns an instance of #{key_str.class}, expected a String"
               end
@@ -330,7 +328,7 @@ module JSON
               first = false
             end
             buf << '}'
-          elsif klass == Array
+          when Array
             buf << '['
             first = true
             obj.each do |e|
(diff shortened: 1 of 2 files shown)


byroot:
> * The only case where an exact class check is needed so far is for String

That's the one case where we had a failing test for it, but does that mean we should special case it?

I'm open to say only String subclasses have that behavior, but then all 3 implementations should match behavior. Could you add some tests to make sure we converge them?

Also I'm very surprised `.class` is this expensive on Truffle, I wouldn't have expected it.

eregon:
I'll rebase and see if the added tests fail on this PR or if they are fine.

eregon:
Tests seem to pass after rebase from a local test run.

byroot (approved):
If tests pass, then that's good for me.

PR #675, opened 2024-11-02
Cache parsed object names

byroot:
Looking up the fstring table isn't that cheap, if we assume most object names are likely to re-appear, it makes sense to keep a small cache of them on the stack.

e.g. `[{"foo": 1, "bar": 2}, {"foo": 3, "bar": 4}]`

In term of implementation, we use a simple sorted array with binary search as it's the most compact and guarantee a decent `O(log n)`, and the comparison is first done on the string length, an then fallback to `memcmp`.

This helps quite a bit on `activitypub` and `twitter` benchmarks, however it is worse on `citm_catalog` because it has a number of unique object names, of the same size, so it behave poorly with the chosen sorting algorithm.

~~It's a bit of a double edge sword, but perhaps a smarter comparison function may solve it.~~ Solved later on.

Before:

```
== Parsing activitypub.json (58160 bytes)
ruby 3.3.4 (2024-07-09 revision be1089c8ec) +YJIT [arm64-darwin23]
Warming up --------------------------------------
                json   714.000 i/100ms
                  oj   801.000 i/100ms
          Oj::Parser   943.000 i/100ms
           rapidjson   632.000 i/100ms
Calculating -------------------------------------
                json      7.135k (± 0.7%) i/s  (140.16 μs/i) -     35.700k in   5.003978s
                  oj      7.991k (± 0.2%) i/s  (125.14 μs/i) -     40.050k in   5.012044s
          Oj::Parser      9.611k (± 0.2%) i/s  (104.04 μs/i) -     48.093k in   5.003723s
           rapidjson      6.318k (± 0.2%) i/s  (158.29 μs/i) -     31.600k in   5.001896s

Comparison:
                json:     7134.7 i/s
          Oj::Parser:     9611.5 i/s - 1.35x  faster
                  oj:     7990.8 i/s - 1.12x  faster
           rapidjson:     6317.6 i/s - 1.13x  slower

== Parsing twitter.json (567916 bytes)
ruby 3.3.4 (2024-07-09 revision be1089c8ec) +YJIT [arm64-darwin23]
Warming up --------------------------------------
                json    57.000 i/100ms
                  oj    62.000 i/100ms
          Oj::Parser    78.000 i/100ms
           rapidjson    56.000 i/100ms
Calculating -------------------------------------
                json    573.527 (± 1.6%) i/s    (1.74 ms/i) -      2.907k in   5.070094s
                  oj    619.368 (± 1.6%) i/s    (1.61 ms/i) -      3.100k in   5.006550s
[...]

PR #676, opened 2024-11-03
JSON::Ext::Parser mark the name cache entries when not on the heap

byroot:
This is somewhat dead code as unless you are using `JSON::Parser.new` direcltly we never allocate `JSON::Ext::Parser` anymore.

But still, we should mark all its reference in case some code out there uses that.

Followup: #675


ext/json/ext/parser/parser.c
```diff
@@ -2408,6 +2408,8 @@ static void JSON_mark(void *ptr)
     rb_gc_mark(json->array_class);
     rb_gc_mark(json->decimal_class);
     rb_gc_mark(json->match_string);
+    const VALUE *name_cache_entries = &json->name_cache.entries[0];
+    rb_gc_mark_locations(name_cache_entries, name_cache_entries + json->name_cache.length);
```
byroot, line 2412:
```
dyld[4099]: missing symbol called
```

cc @eregon looks like `rb_gc_mark_locations` is also missing on Truffle?

PR #679, opened 2024-11-04
Add tests for the behavior of JSON.generate with base types subclasses

casperisfine:
Ref: https://github.com/ruby/json/pull/674
Ref: https://github.com/ruby/json/pull/668

The behavior on such case it quite unclear, the goal here is to figure out whatever was the behavior on Cext version of `json 2.7.0` and get all implementations to converge.

We can then decide to make them all behave differently if we so wish.

FYI: @eregon


test/json/json_generator_test.rb
```diff
@@ -536,6 +536,72 @@ def test_to_json_called_with_state_object
...
+    assert_equal '{"[]":1}', JSON.generate(ArrayWithToJSON.new => 1)
+  end
+
+  class ArrayWithToS < Array
+    include CustomToS
+  end
+
+  def test_array_subclass_with_to_s
+    assert_equal '[]', JSON.generate(ArrayWithToS.new)
+    assert_equal '{"JSONGeneratorTest::ArrayWithToS#to_s":1}', JSON.generate(ArrayWithToS.new => 1)
```
eregon, line 566: This doesn't seem reasonable, i.e. calling `to_s` on an Array subclass (both semantically and performance-wise).
Inheriting from core classes is frowned upon, so without a convincing use case I see no value to respect those old weird/broken semantics.

eregon:
For the ActiveSupport use case in https://github.com/ruby/json/issues/667 maybe this extra escaping should/could be a JSON.dump option? Then maybe we wouldn't need to treat subclasses any specially.

casperisfine:
I created a self contained test case to more easily test historical behavior, and it seems we already diverged: https://gist.github.com/casperisfine/a379245862b0d5273086e5053657460a

I'll update the test suite to have the same behavior as 2.7.2

ext/json/ext/generator/generator.c
```diff
@@ -755,7 +755,11 @@ json_object_i(VALUE key, VALUE val, VALUE _arg)
     VALUE key_to_s;
     switch(rb_type(key)) {
         case T_STRING:
-            key_to_s = key;
+            if (RB_LIKELY(RBASIC_CLASS(key) == rb_cString)) {
+                key_to_s = key;
+            } else {
+                key_to_s = rb_funcall(key, i_to_s, 0);
+            }
```
casperisfine, line 762: We had regressed here on the C implementation.

test/json/json_generator_test.rb
eregon: I think ideally we wouldn't treat subclasses specially at all, it doesn't seem reasonable to subclass a core type and then expect different behavior than the superclass for JSON behavior and many other things (Liskov substitution principle).
If someone wants to have some custom serialization to JSON, it seems simple enough to have a custom type not inheriting from a core type and having `to_json` implemented on it.

casperisfine: I'm not disagreeing, I'm just saying we should be very deliberate about it, because backward compatibility is very important.

eregon: Ah for this specific example I missed that `to_s` is only called because the `ArrayWithToS` is a Hash key, that seems less crazy at least :sweat_smile:

Issue #650, opened 2024-10-25
Figure out what to do when different versions of `json` and `json_pure` are loaded


byroot:
Ref: https://github.com/ruby/json/issues/646

There is no version constraints between the two, but they expose the same files in `$LOAD_PATH`, so if two different versions are loaded it can easily cause incompatibilities.

We should find a solution for this. Perhaps use `require_relative`? But even then, large parts of the implementation are in the global `JSON` module, so we can't load two namespaced versions of it.

Earlopain:
Since this is an unsupported use-case, can they both define a version constant somewhere independent from each other?

```rb
begin
  require "json/pure/version"
  require "json/ext/version"
  if JSON::Ext::VERSION != JSON::Pure::VERSION
    raise "Some descriptive error"
  end
rescue LoadError
end
```

This leaves people with older versions out of luck but maybe its acceptable? OTOH, loading different versions probably has just worked for a long time.

byroot:
Yet another such issue https://github.com/ruby/json/pull/652

> `if JSON::Ext::VERSION != JSON::Pure::FromGem::VERSION`

Yeah, this is kinda the obvious / easy thing to do. But I wonder if it's really what should be done.

I also wonder if the solution isn't simply to deprecate and remove `json_pure`.

After all, now that `json` is in the stdlib, why would you need `json_pure` specifically?

byroot:
When I have some time, I'd like to go over look more into https://rubygems.org/gems/json_pure/reverse_dependencies, because at first sight I don't see a lot of legitimate use.

So I think we could just push a `json_pure` that's just a `gemspec` that require `json`. The `Pure` implementation would remain inside `json` but we'd enforce that only one implementation is ever loaded (e.g. use the pure generator on truffle, etc).

headius:
> why would you need json_pure specifically?

Didn't we just have an issue where someone wanted to use json_pure because they deployed on a system without build tools?

byroot:
It was resolved by the fact that `json` is part of the stdlib, so there's no need to build it.

eregon:
> It was resolved by the fact that `json` is part of the stdlib, so there's no need to build it.

Mmh, but that requires to use the same version of json as in stdlib (which might not be an option if some bug fix or change is wanted) and recent Bundler always reinstall default gems (json is a default gem): https://github.com/rubygems/rubygems/issues/8180

[...]

PR #682, opened 2024-11-05
Prevent loading both `json/ext` and `json/pure`

casperisfine:
Fix: https://github.com/ruby/json/issues/650
Ref: https://github.com/ruby/json/issues/646

`json_pure` currently doesn't work well at all because `json/ext` is a default gem, so if you have code requiring `json/pure` and some other code requiring `json`, you end up with both loaded.

If the `json` and `json_pure` versions match, it's not too bad, but if they don't match, you might run into issues with private APIs no longer matching.

### DIFF
--- a/json_pure.gemspec
+++ b/json_pure.gemspec
@@ -16,6 +16,7 @@ Gem::Specification.new do |s|
 
   s.extra_rdoc_files = ["README.md"]
   s.rdoc_options = ["--title", "JSON implementation for ruby", "--main", "README.md"]
+
   s.files = [
     "CHANGES.md",
     "COPYING",
@@ -23,29 +24,9 @@ Gem::Specification.new do |s|
     "LEGAL",
     "README.md",
     "json_pure.gemspec",
-    "lib/json.rb",
-    "lib/json/add/bigdecimal.rb",
-    "lib/json/add/complex.rb",
-    "lib/json/add/core.rb",
-    "lib/json/add/date.rb",
-    "lib/json/add/date_time.rb",
-    "lib/json/add/exception.rb",
-    "lib/json/add/ostruct.rb",
-    "lib/json/add/range.rb",
-    "lib/json/add/rational.rb",
-    "lib/json/add/regexp.rb",
-    "lib/json/add/set.rb",
-    "lib/json/add/struct.rb",
-    "lib/json/add/symbol.rb",
-    "lib/json/add/time.rb",
-    "lib/json/common.rb",
-    "lib/json/ext.rb",
-    "lib/json/generic_object.rb",
-    "lib/json/pure.rb",
-    "lib/json/pure/generator.rb",
-    "lib/json/pure/parser.rb",
-    "lib/json/version.rb",
+    *(Dir["lib/**/*.rb"] - Dir["lib/json/ext/**/*.rb"]),
   ]
+
   s.homepage = "https://ruby.github.io/json"
   s.metadata = {
     'bug_tracker_uri'   => 'https://github.com/ruby/json/issues',
--- a/lib/json/ext.rb
+++ b/lib/json/ext.rb
@@ -1,25 +1,30 @@
 # frozen_string_literal: true
 
-require 'json/common'
+unless defined?(::JSON::JSON_LOADED)
+  module JSON
+    # Prevent json/pure from being loaded at the same time
+    $LOADED_FEATURES << File.expand_path("../pure.rb", __FILE__)
 
-module JSON
-  # This module holds all the modules/classes that implement JSON's
-  # functionality as C extensions.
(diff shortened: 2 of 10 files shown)


lib/json/ext.rb
```diff
@@ -2,24 +2,26 @@
...
-    else
-      require 'json/ext/parser'
-      require 'json/ext/generator'
-      $DEBUG and warn "Using Ext extension for JSON."
-      JSON.parser = Parser
-      JSON.generator = Generator
-    end
+unless defined?(::JSON::JSON_LOADED)
+  # Prevent json/pure from being loaded at the same time
+  $LOADED_FEATURES << File.expand_path("../pure.rb", __FILE__)
```
casperisfine, line 7: I tired to add some stub `json/ext/generator.rb` etc, but it was making loading the test suite very contrived.

In the end I think marking the alternative implementation as loaded is OK?

lib/json_pure.rb
```diff
@@ -0,0 +1,3 @@
+# frozen_string_literal: true
+
+require 'json/pure'
```
casperisfine, line 3: This is a bit orthogonal, but ensure that adding `gem "json_pure"` in a Gemfile properly load `json/pure` in the expected order.

json_pure.gemspec
```diff
@@ -16,36 +16,27 @@ Gem::Specification.new do |s|
...
+    "CHANGES.md",
+    "COPYING",
+    "BSDL",
+    "LEGAL",
+    "README.md",
+    "json.gemspec",
+    *(Dir["lib/**/*.rb"] - Dir["lib/json/ext/**/*.rb"]),
+  ]
+
   s.files = [
```
eregon, line 30: You're setting the files twice here

casperisfine: Good catch. Fixed.

json_pure.gemspec
```diff
@@ -16,36 +16,27 @@ Gem::Specification.new do |s|
...
   s.rdoc_options = ["--title", "JSON implementation for ruby", "--main", "README.md"]
+
+  s.files = [
+    "CHANGES.md",
+    "COPYING",
+    "BSDL",
+    "LEGAL",
+    "README.md",
+    "json.gemspec",
+    *(Dir["lib/**/*.rb"] - Dir["lib/json/ext/**/*.rb"]),
```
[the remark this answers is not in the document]
casperisfine: >  as you could get a mix of which files get loaded from where, especially if different versions.

That's the opposite. This way if you got `json_pure` above `json` in your gemfile, it takes precedence.

The goal is exactly for both gems to define these "features" (as in `$LOADED_FEATURES`) so we prevent mixed loading.

> IMO if one's environment is too broken to install extension gems, then they simply can't install it, and I don't think json needs to workaround such broken environments.

Some people don't want compilers in production, but yes, IMO the value of `json_pure` is really doubtful now that `json` is a default gem, if you can't compile extensions, just figure out how to use the stdlib.

But in the meantime I think this PR is a decent way forward.

json_pure.gemspec
```diff
@@ -16,36 +16,17 @@ Gem::Specification.new do |s|
...
-    "lib/json/add/symbol.rb",
-    "lib/json/add/time.rb",
-    "lib/json/common.rb",
-    "lib/json/ext.rb",
-    "lib/json/generic_object.rb",
-    "lib/json/pure.rb",
-    "lib/json/pure/generator.rb",
-    "lib/json/pure/parser.rb",
-    "lib/json/version.rb",
+    "json.gemspec",
```
eregon, line 26:
```suggestion
    "json_pure.gemspec",
```
or not including it at all

lib/json/ext.rb
eregon: Given that file doesn't exist (because of the changes in json.gemspec), does it have any effect?
It feels brittle.
Is it necessary? I would think the `JSON::JSON_LOADED` check is enough.

eregon:
Also in that scenario `lib/json/common.rb` would be loaded from json_pure, but maybe there is some code there which is not compatible with what the C extension expects.

lib/json/ext.rb
casperisfine: It's in case the version of the "other gem" is much older.

casperisfine:
Hum, right, I think I need to use more `require_relative`.

eregon:
I think we should just have `json_pure` an empty gem depending on `json`, it seems intrinsically very brittle to have 2 gems which share some Ruby files (e.g. `lib/json/common.rb` and `lib/json/version.rb`) of potentially different versions.

If we wanted to just make sure the versions match we'd need to have json_pure depend on json same version, which is the same thing and then I think there is no point to ship the same files in two gems.

PR #685, opened 2024-11-05
Turn `json_pure` into an empty gem

casperisfine:
Fix: https://github.com/ruby/json/issues/650
Closes: https://github.com/ruby/json/pull/682

cc @eregon @earlopain


eregon:
I didn't realize this would mean removing `lib/json/pure/parser.rb` but I think it makes sense, no point to maintain another backend if it's not used in any case.

We could have kept `require 'json/pure'` to require the pure generator & parser, but I suppose there is very limited value/use cases for that.

Issue #524, opened 2023-03-31
Streaming enhancements for dumping


headius:
While investigating workarounds for https://github.com/jruby/jruby/issues/6265 I realized that all dumping for e.g. `to_json` is done first to an in-memory buffer (always a Ruby String) even when given an IO object to which the json should be written. This applies to all three implementations: the pure-Ruby version, the C version, and the Java version.

This could obviously be more efficient if the json appends were writes directly to the given IO, or if it were possible to provide a String-like object that receives the appends. A rework of the generator subsystem would be necessary to pass any provided IO or String-like through the various dump methods.

This would have several benefits:

* No intermediate String to hold the entirety of the dumped json.
* No intermediate Strings for components of a dumped collection; Array and Hash currently dump each element or pair to a separate String and then append that String to the result buffer.
* Reduced allocation, copying, and GC overhead when dumping directly to IO.
* Potential to provide IO-like or String-like receivers of the dumped json, allowing for a workaround to the Java 2GB array limitation (https://github.com/jruby/jruby/issues/6265).

I'm hoping to attempt this for at least the Java and Ruby versions of the generator, but I may need help making the same change in the C extension. If others are interested in helping with any of these implementations, it would be greatly appreciated.

PR #686, opened 2024-11-05
JSON.dump: write directly into the provided IO

byroot:
Ref: https://github.com/ruby/json/issues/524

Fix: https://github.com/ruby/json/pull/538
Fix: https://github.com/ruby/json/issues/524

Rather than to buffer everything in memory.

Unfortunately Ruby doesn't provide an API to write into and IO without first allocating a string, which is a bit wasteful.

FYI: @headius

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -161,7 +161,7 @@ if defined?(RUBY_ENGINE) and RUBY_ENGINE == 'jruby'
   file JRUBY_PARSER_JAR => :compile do
     cd 'java/src' do
       parser_classes = FileList[
-        "json/ext/ByteListTranscoder*.class",
+        "json/ext/ByteList*.class",
         "json/ext/OptionsReader*.class",
         "json/ext/Parser*.class",
         "json/ext/RuntimeInfo*.class",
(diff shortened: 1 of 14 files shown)


headius:
Java side is implemented atop this PR in #538.

headius:
We need some benchmarks!

byroot:
I cherry picked you commits.

Now TruffleRuby head seem to crash for some reason, but I'm having trouble understanding Truffle crash reports, so I'm going to assume it's just a bug in Truffle. FYI: @eregon 

Also the `TruffleRuby::Generator` is still buffering everything in a string before flushing, like other implementations used to.

headius:
@byroot Thanks! I have some optimizations I'll do in a separate PR now.

andrykonchin:
@eregon The issue is fixed in https://github.com/oracle/truffleruby/commit/567aa2763b79c54e68615e66d8d4221cadb871d4.

byroot:
Urk, I didn't realize `rb_gc_mark_locations` was using `rb_gc_mark_maybe` under the hood. I probably should use a loop instead.

eregon:
Yes, I've never seen `rb_gc_mark_locations()` before, I think a simple loop would be better/safer and as fast or faster.

PR #688, opened 2024-11-06
Add changes for v2.7.6

y-yagi:
I picked from https://github.com/ruby/json/blob/v2.7.x/CHANGES.md


byroot:
I was planing to do it soon, but thank you!

PR #691, opened 2024-11-06
Move extension directories under `ext/json`

nobu:
`ext/json/ext` feels redundant.


byroot:
> `ext/json/ext` feels redundant.

It definitely is, and I do indeed what to simplify all that, but I'm a bit worried about potential backward compatibility issues, so I may delay this for 3.0.

nobu:
I found the current `Rakefile` is evil of evil.
Must rewrite it first.

PR #695, opened 2024-11-06
CI: ci.yml - update to actions/checkout@v4

MSP-Greg:
Fixes Actions warning like:
```text
macos-14 3.2 The following actions use a deprecated Node.js version and will be forced to run on node20: actions/checkout@v3.
For more info: https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default
Show more
```


byroot (approved):
❤️

Issue #710, opened 2024-11-24
Better error debugging for generation


mperham:
I'm trying to figure out why JSON is blowing up when generating the JSON for a big ole Hash of stuff. It raises a GeneratorError, but with no context so I can't easily tell where my bad data is. The line in question is here:

https://github.com/ruby/json/blob/55015fa91eb48a12bebe33869b7048b3e9818154/ext/json/ext/generator/generator.c#L912

I would suggest this error include the relevant object at fault. Something like:

```ruby
class JSON::GeneratorError
  attr_reader :invalid_element
end
```

byroot:
Makes sense. Wouldn't be too hard to implement.

mperham:
Thank you, it worked great.

Should it emit the invalid object as part of `message`? That's extremely useful context that most people won't know exists. Is the attribute in the RDoc?

byroot:
The object may be huge, that's why it's included in `detailed_message` and not just `message`, to avoid things like: https://bugs.ruby-lang.org/issues/18285

mperham:
TIL about `#detailed_message`! I think I could roll this into Sidekiq's debugging output: turn on debugging => get detailed_message logged which will solve my need.

https://docs.ruby-lang.org/en/master/Exception.html#method-i-detailed_message

PR #712, opened 2024-11-25
JSON::GeneratorError expose invalid object

byroot:
Fix: https://github.com/ruby/json/issues/710

Makes it easier to debug why a given tree of objects can't be dumped as JSON.

### DIFF
--- a/lib/json/common.rb
+++ b/lib/json/common.rb
@@ -143,7 +143,23 @@ module JSON
   # :startdoc:
 
   # This exception is raised if a generator or unparser error occurs.
-  class GeneratorError < JSONError; end
+  class GeneratorError < JSONError
+    attr_reader :invalid_object
+
+    def initialize(message, invalid_object = nil)
+      super(message)
+      @invalid_object = invalid_object
+    end
+
+    def detailed_message(...)
+      if @invalid_object.nil?
+        super
+      else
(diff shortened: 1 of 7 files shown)


lib/json/common.rb
```diff
@@ -143,7 +143,23 @@ class CircularDatastructure < NestingError; end
...
+    def initialize(message, invalid_object = nil)
+      super(message)
+      @invalid_object = invalid_object
+    end
+
+    def detailed_message(...)
+      if @invalid_object.nil?
+        super
+      else
+        "#{super}\nInvalid object: #{@invalid_object.inspect}"
```
mperham, line 158: The only quibble I've run into when pulling this into Sidekiq is the newline here. It's injecting formatting into the message, typically if I want to log an error message I don't expect it to take multiple lines in the log output. WDYT?

mperham: Of course we expect newlines with backtraces... Could exceptions support a hash of context values to be logged with the error, where `invalid_object` would already be there? This is pretty standard stuff for structured logging, e.g. [Golang](https://pkg.go.dev/log/slog#Logger.With).

```ruby
rescue JSON::GeneratorError => ex
  logger.info { ex.with_context(some: value, another: value).detailed_message }
```

I guess the bigger question would be if ruby-core wants to move toward more formalized structured logging, by providing these types of APIs.

PR #718, opened 2024-12-16
Introduce JSON::Coder

etiennebarrie:
Since `#to_json` methods are global, it can sometimes be problematic if you need a given type to be
serialized in different ways in different locations.

Instead it is recommended to use the newer `JSON::Coder` API:

```ruby
module MyApp
  API_JSON_CODER = JSON::Coder.new do |object|
    case object
    when Time
      object.iso8601(3)
    else
      object
    end
  end
end

puts MyApp::API_JSON_CODER.dump(Time.now.utc) # => "2025-01-21T08:41:44.286Z"
```

The provided block is called for all objects that don't have a native JSON equivalent, and
must return a Ruby object that has a native JSON equivalent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ If bundler is not being used to manage dependencies, install the gem by executin
 
     $ gem install json
 
-## Usage
+## Basic Usage
 
 To use JSON you can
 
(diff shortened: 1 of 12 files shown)


paddor:
I love this. That's exactly what I'd need to make sure my `Time` objects are encoded as ISO8601 with millisecond precision without monkey patching.

Doing this just doesn't feel right:

```ruby
class Time
  def to_json(*a)
    iso8601(3).to_json(*a)
  end
end
```

So far I've been relying on Oj's `time_format: :xmlschema, second_precision: 3`.

PR #719, opened 2024-12-18
Add support for Solaris 10 which lacks strnlen()

ngoto:
json parser fails to start on Solaris 10 because Solaris 10 lacks strnlen().

To add support for Solaris 10, added a check for the existence of strnlen() in ext/json/ext/parser/extconf.rb, and added alternative strnlen() implementation to be used if it is missing.


byroot:
Thank you. I'll merge on green and cut a 2.9.1 release.

PR #725, opened 2025-01-09
Apply recent C optimizations to Java encoder

headius:
Just catching up with all of @byroot's excellent optimization work.


headius:
As part of this I'll be trying to align more of the Java code with the C equivalents, with comments to indicate how they sync up. This should make it easier to keep them in sync in the future.

headius:
I jumped down the optimization well and continued past the recent string optimizations on to the other types of dumpable objects. Strings are now faster than in CRuby, as well as a few other cases in the encoder.rb benchmark, but many cases are still slower... sometimes less than half the performance.

Tracking results here: https://gist.github.com/headius/3e56d80656543bf2343f4b26f00bc446

headius:
A nice discovery: the default ByteArrayOutputStream we extend for our ByteList version uses `synchronized` on its write methods, and removing that (by avoiding the JDK impl) improves performance of dumping substantially.

headius:
In the interest of getting this part merged I've narrowed scope to the encoder. I'll need to incorporate a handful of recent changes, but they're not too extensive.

byroot:
I've pre-merged your changes to the benchmark, so it should be easier for you to rebase.

Feel free to merge whenever you fell like it's ready, as previously said, as far as I'm concerned, you are the main maintainer of the Java parts.

Please just try to have a clean git history before merging.

headius:
> Please just try to have a clean git history before merging.

What do you mean by "clean git history"?

byroot:
> What do you mean by "clean git history"?

Well for instance in 4d37e9fee12fa3f8cb819561729afcdb2e9cc348 you made some changes and forgot to remove some unused imports, and then 3 commits later on there is 98cb7859c71a78a9ee2557f29fccb69d83e1acac which remove the unused import.

Both of this should be a single commit really. 

But if you don't care about the git history, just squash it all I guess.

headius:
I care about the git history and don't want to squash, but I'm not going to rewrite 24 commits just to regroup cleanup changes. Import cleanup (like code reformatting or whitespace changes) has no effect on the build or on run-time functionality. I always try to keep things clean but this will happen from time to time.

byroot:
> I'm not going to rewrite 24 commits just to regroup cleanup changes.

Then squash.

headius:
I have merged in recent changes from master.

@etiennebarrie Could you review to make sure I properly kept your changes? Tests pass but I want have another set of eyes on it.

byroot:
> @etiennebarrie Could you review to make sure I properly kept your changes? 

Looks fine to me.

etiennebarrie:
Yes, the important thing is the fix to JSON::Coder to allow special handling of `Float::NAN` and `Float::INFINITY`, it seems to work fine:
```console
$ ruby -v -Ilib:ext -rjson -e 'puts JSON::Coder.new { |value| raise unless [Float::INFINITY, Float::NAN].include? value }.dump(a: Float::NAN)'
jruby 9.4.11.0 (3.1.4) 2025-01-29 9b107851a3 OpenJDK 64-Bit Server VM 23.0.2 on 23.0.2 +jit [arm64-darwin]
{"a":null}
```

headius:
> Yes, the important thing is the fix to JSON::Coder to allow special handling of `Float::NAN` and `Float::INFINITY`, it seems to work fine:

Thank you! I almost missed the Float changes but the tests caught my mistake.

PR #733, opened 2025-01-20
Fix parsing incomplete unicode escape "\uaaa"

tompng:
Fix segmentation fault parsing `"\uaaa"`
```ruby
1000.times { JSON.parse '"\\uaaa"' rescue nil }
```


byroot (approved):
❤️

PR #734, opened 2025-01-20
Raise parse error on invalid comments

tompng:
Fix parsing these invalid json
```ruby
JSON.parse '1 /*aaa' # => 1, should be error
JSON.parse '1 /' # => never stops
JSON.parse '[1, /' # => never stops
```


byroot (approved):
❤️

PR #742, opened 2025-01-31
Refactor convert_UTF8_to_JSON to split searching and escaping code

byroot:
The goal is to be able to dispatch to more optimized search implementations without having to duplicate the escaping code.

Somehow, this is a few % faster already:

```
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin23]
               after     2.257k i/100ms
               after     22.930k (± 1.3%) i/s   (43.61 μs/i) -    115.107k in   5.020814s

Comparison:
              before:    21604.0 i/s
               after:    22930.1 i/s - 1.06x  faster

ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin23]
               after   137.000 i/100ms
               after      1.397k (± 1.1%) i/s  (715.57 μs/i) -      6.987k in   5.000408s

Comparison:
              before:     1344.4 i/s
               after:     1397.5 i/s - 1.04x  faster

ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +YJIT +PRISM [arm64-darwin23]
               after   249.000 i/100ms
               after      2.464k (± 1.8%) i/s  (405.81 μs/i) -     12.450k in   5.054131s

Comparison:
              before:     2326.5 i/s
               after:     2464.2 i/s - 1.06x  faster
```

cc @samyron I think that should make integrating SIMD and bit twiddling algorithms easier.

The existence of a state struct should also allow to resume search without rescaning the same bytes and stay aligned like https://lemire.me/blog/2024/07/20/scan-html-even-faster-with-simd-instructions-c-and-c/

What do you think? Does this new refactoring makes things easier for you? (I think it does, but maybe you have a different opinion).

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -101,6 +101,7 @@ static void raise_generator_error(VALUE invalid_object, const char *fmt, ...)
 // 0 - single byte char that don't need to be escaped.
 // (x | 8) - char that needs to be escaped.
 static const unsigned char CHAR_LENGTH_MASK = 7;
+static const unsigned char ESCAPE_MASK = 8;
 
 static const unsigned char escape_table[256] = {
     // ASCII Control Characters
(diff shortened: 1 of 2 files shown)


ext/json/ext/generator/generator.c
```diff
@@ -165,6 +166,84 @@ static const unsigned char script_safe_escape_table[256] = {
...
+        if (RB_UNLIKELY(ch_len)) {
+            if (ch_len & ESCAPE_MASK) {
+                if (RB_UNLIKELY(ch_len == 11)) {
+                    const unsigned char *uptr = (const unsigned char *)state->ptr;
+                    if (!(uptr[1] == 0x80 && (uptr[2] >> 1) == 0x54)) {
+                        state->ptr += 3;
+                        continue;
+                    }
+                }
+                search_flush(state);
```
byroot, line 234: Since `search_escape` is now responsible for flushing bytes that don't need escaping, it should allow to write an SIMD version that does both scanning and copying at once. e.g. load X bytes, and if none need to be escaped, directly copy them.

samyron:
I think this looks great. The separation between the searching and the escaping should make the SIMD integration easier. 

After this gets merged, I'll re-work the SIMD branches. I might close the existing PRs and create smaller more targeted PRs for each SIMD implementation.

PR #745, opened 2025-02-05
Optimize Symbol generation in strict mode

etiennebarrie:
In strict mode we don't need to check for `to_json`, we can just generate a String from a Symbol.

It changes behavior because previously Symbol caused a generator error. But keys in hashes already had support for automatic coercion to String, so it feels ok to do.

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -991,6 +991,29 @@ static void generate_json_string(FBuffer *buffer, struct generate_json_data *dat
     fbuffer_append_char(buffer, '"');
 }
 
+static void generate_json_fallback(FBuffer *buffer, struct generate_json_data *data, JSON_Generator_State *state, VALUE obj)
+{
+    VALUE tmp;
+    if (rb_respond_to(obj, i_to_json)) {
+        tmp = rb_funcall(obj, i_to_json, 1, vstate_get(data));
+        Check_Type(tmp, T_STRING);
+        fbuffer_append_str(buffer, tmp);
+    } else {
+        tmp = rb_funcall(obj, i_to_s, 0);
+        Check_Type(tmp, T_STRING);
+        generate_json_string(buffer, data, state, tmp);
+    }
+}
+
+static inline void generate_json_symbol(FBuffer *buffer, struct generate_json_data *data, JSON_Generator_State *state, VALUE obj)
+{
+    if (state->strict) {
+        generate_json_string(buffer, data, state, rb_sym2str(obj));
+    } else {
+        generate_json_fallback(buffer, data, state, obj);
+    }
+}
+
 static void generate_json_null(FBuffer *buffer, struct generate_json_data *data, JSON_Generator_State *state, VALUE obj)
 {
     fbuffer_append(buffer, "null", 4);
@@ -1049,7 +1072,6 @@ static void generate_json_fragment(FBuffer *buffer, struct generate_json_data *d
 
 static void generate_json(FBuffer *buffer, struct generate_json_data *data, JSON_Generator_State *state, VALUE obj)
 {
-    VALUE tmp;
     bool as_json_called = false;
 start:
     if (obj == Qnil) {
@@ -1063,6 +1085,8 @@ start:
             generate_json_fixnum(buffer, data, state, obj);
         } else if (RB_FLONUM_P(obj)) {
             generate_json_float(buffer, data, state, obj);
+        } else if (RB_STATIC_SYM_P(obj)) {
+            generate_json_symbol(buffer, data, state, obj);
         } else {
             goto general;
         }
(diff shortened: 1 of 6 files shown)


ext/json/ext/generator/generator.c
```diff
@@ -1084,6 +1108,10 @@ static void generate_json(FBuffer *buffer, struct generate_json_data *data, JSON
                 if (klass != rb_cString) goto general;
                 generate_json_string(buffer, data, state, obj);
                 break;
+            case T_SYMBOL:
+                if (klass != rb_cSymbol) goto general;
+                generate_json_string(buffer, data, state, rb_sym2str(obj));
```
eregon, line 1113: Shouldn't it be `generate_json_symbol()` too?
Or at least I think static & dynamic symbols should be treated consistently.

ext/json/ext/generator/generator.c
```diff
@@ -1084,6 +1108,10 @@ static void generate_json(FBuffer *buffer, struct generate_json_data *data, JSON
                 if (klass != rb_cString) goto general;
                 generate_json_string(buffer, data, state, obj);
                 break;
+            case T_SYMBOL:
+                if (klass != rb_cSymbol) goto general;
```
eregon, line 1112: This should be impossible:
```
irb(main):002> Class.new(Symbol).new
(irb):2:in `<main>': undefined method `new' for class #<Class:0x00007efe377b26e0> (NoMethodError)
irb(main):004> c=Class.new(Symbol)
=> #<Class:0x00007efe377b4f80>
irb(main):006> Class.instance_method(:new).bind_call(c)
(irb):6:in `new': allocator undefined for #<Class:0x00007efe377b4f80> (TypeError)

Class.instance_method(:new).bind_call(c)
                                      ^
	from (irb):6:in `bind_call'
```

ext/json/ext/generator/generator.c
byroot: Yes it should. It's a small mistake. I need to go over the JRuby and TruffleRuby implementations anyway.

ext/json/ext/generator/generator.c
```diff
@@ -991,6 +991,29 @@ static void generate_json_string(FBuffer *buffer, struct generate_json_data *dat
...
+    } else {
+        tmp = rb_funcall(obj, i_to_s, 0);
+        Check_Type(tmp, T_STRING);
+        generate_json_string(buffer, data, state, tmp);
+    }
+}
+
+static inline void generate_json_symbol(FBuffer *buffer, struct generate_json_data *data, JSON_Generator_State *state, VALUE obj)
+{
+    if (state->strict) {
```
eregon, line 1010: It seems unfortunate to only do this fast path when strict is set, should we maybe always do it?

lib/json/truffle_ruby/generator.rb
```diff
@@ -303,7 +303,7 @@ def to_h
         # GeneratorError exception.
         def generate(obj, anIO = nil)
           if @indent.empty? and @space.empty? and @space_before.empty? and @object_nl.empty? and @array_nl.empty? and
-              !@ascii_only and !@script_safe and @max_nesting == 0 and !@strict
```
eregon, line 306: I think we still need this check, the `else` of the `case` doesn't check for `@strict`

test/json/json_generator_test.rb
```diff
@@ -86,6 +86,9 @@ def test_dump_strict
 
     assert_equal '42', dump(42, strict: true)
     assert_equal 'true', dump(true, strict: true)
+
+    assert_equal '"hello"', dump(:hello, strict: true)
+    assert_equal '"World"', dump("World".to_sym, strict: true)
```
eregon, line 91: Probably should test that `to_json` is not called with strict

ext/json/ext/generator/generator.c
eregon: i.e. I doubt anyone relies on `Symbol#to_json` to be called, there are not that many ways to dump a Symbol to JSON

byroot: Unfortunately we can't, because of this stupid `json/add/symbol`: https://github.com/ruby/json/blob/c472d72fdf1fd8e297cfdaf05b95b1ddcfef63fb/lib/json/add/symbol.rb

test/json/json_generator_test.rb
byroot: It's somewhat already asserting that because `json/add/symbol` is loaded in the test suite.

eregon:
FWIW I realized this means `JSON.load(JSON.dump(obj, strict: true), strict: true)` does no longer always round-trip cleanly, but I'm not sure how much that matters.

byroot:
I think that ship has sailed when symbols were accepted as hash keys regardless.

But yes, we could chose to do the opposite and reject them in strict mode. One goal of `strict: true` was indeed to help with roundtripping :/

PR #753, opened 2025-02-14
Pass through all options if present.

ioquatix:
I understand there is a compatibility concern here https://github.com/ruby/json/commit/8e0076a3f2f8a7277e659607ab195845ce505a01

However, I'd like to try this out (check CI is passing).

### DIFF
--- a/lib/json/common.rb
+++ b/lib/json/common.rb
@@ -840,7 +840,7 @@ module JSON
 
     opts = JSON.dump_default_options
     opts = opts.merge(:max_nesting => limit) if limit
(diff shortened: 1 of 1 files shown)


ioquatix:
For reference, it's also possible to achieve what I want using `JSON::State.generate` directly:

```
OPTIONS = {indent: "  ", space: " ", space_before: "", object_nl: "\n", array_nl: "\n"}

def self.output(file, value)
  ::JSON::State.generate(value, OPTIONS, file)
  file.puts
end
```

PR #759, opened 2025-03-10
Avoid fast-path IO writes when IO has ext enc

headius:
This works around a bug in JRuby's IOOutputStream logic whereby an IO with an external encoding will always fail to write incoming bytes. We use base logic to detect if the target object is a "real IO" and if so and it has an external encoding, we drop the realIO and. This causes the rest of IOOutputStream to avoid the fast path and always use dyncall logic with RubyString wrappers.

This works around the issue in jruby/jruby#8682.

This should be temporary, since it will definitely degrade direct writes to such IO objects, but a longer-term fix for the encoding issues spelled out in jruby/jruby#8682 will need to come first, or else json will have to be modified to not use IOOutputStream at all.

### DIFF
--- a/java/src/json/ext/Generator.java
+++ b/java/src/json/ext/Generator.java
@@ -5,6 +5,7 @@
  */
 package json.ext;
 
+import org.jcodings.Encoding;
 import org.jcodings.specific.UTF8Encoding;
 import org.jruby.Ruby;
 import org.jruby.RubyArray;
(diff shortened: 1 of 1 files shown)


headius:
The root issue this works around is technically jruby/jruby#6588 but this temporary fix will address the failing rdoc test from jruby/jruby#8682 and allow the test to be restored.

headius:
#760 details the ongoing work needed after this workaround.

headius:
@byroot @nobu When this is released, the rdoc test disabled for jruby/jruby#8682 can be restored by either getting a new JRuby out or by installing latest json gem.

PR #762, opened 2025-03-11
Raise a ParserError on all incomplete unicode escape sequence.

byroot:
This was the behavior until `2.10.0` unadvertently changed it.

`"\u1"` would raise, but `"\u1zzz"` wouldn't.


byroot:
Not too sure what's wrong with ruby-head.

Earlopain:
Something in optparse https://github.com/ruby/optparse/issues/93 with rake and environment variables

PR #767, opened 2025-03-16
Faster integer formatting

radiospiel:
This PR provides an alternative implementation for a long → decimal conversion.

The main difference is that it uses an algorithm pulled from https://github.com/jeaiii/itoa. The source there is C++, it was converted by hand to C for inclusion with this gem. A writeup of this algorithm can be found [here](https://jk-jeon.github.io/posts/2022/02/jeaiii-algorithm/). jeaiii's algorithm is covered by the MIT License, see source code.

On addition this version now also generates the string directly into the `fbuffer`, foregoing the need to run a separate memory copy.

As a result, I see a speedup of 32% on Apple Silicon M1 for an integer set of benchmarks. 

```
== Encoding ints (45025 bytes)
        json (local)     1.566k i/100ms
        json (local)     16.322k (± 3.9%) i/s   (61.27 μs/i) -     82.998k in   5.093499s

Normalize to 8009 byte
== Encoding ints (45025 bytes)
       json (2.10.2)     1.228k i/100ms
       json (2.10.2)     12.352k (± 0.5%) i/s   (80.96 μs/i) -     62.628k in   5.070472s
```

The script adds a "cents" test case ([link](https://gist.github.com/radiospiel/04019402726a28b31616df3d0c17bd1c#file-encoder-rb-L188)) which explicitly covers integers in a range that is relevant for financial institutions. (Comparing benchmarks for that case show a speedup of 41%)

The `dumps` benchmark, which covers the JSON files in `benchmark/data/*.json`, sees a speedup of ~8 %.

### DIFF
--- a/ext/json/ext/fbuffer/fbuffer.h
+++ b/ext/json/ext/fbuffer/fbuffer.h
@@ -3,6 +3,7 @@
 
 #include "ruby.h"
 #include "ruby/encoding.h"
+#include "../vendor/jeaiii-ltoa.h"
 
 /* shims */
 /* This is the fallback definition from Ruby 3.4 */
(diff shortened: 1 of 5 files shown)


radiospiel:
Note: see [this](https://github.com/ruby/json/pull/768#issuecomment-2728622581) for discussing licensing concerns. Marking this PR as draft for the time being.

radiospiel:
The windows build errors have been fixed. This PR is ready for review.

The changes are most impactful for long integers. A tests on 19 digit numbers shows a 65% speedup; the "cents" test (a mix of numbers between 1 and 7 digits) has 41%. On 2-digit numbers the gains are still  ~11%. (all tested on macOS, Apple Silicon M1)

I think a 32-bit version could be built and might over some small gains. The C++ version, which is the base for these changes, specialises templates for various integer widths. I didn't want to go /that/ far here.

PR #769, opened 2025-03-16
Implement JSON string escaping using SIMD (ARM + X86)

radiospiel:
It integrates a `simd.h` shim extracted from Postgresql ([src](https://github.com/postgres/postgres/blob/REL_17_4/src/include/port/simd.h)) Postgresql is licensed under a MIT/BSD-style license ([link](https://github.com/postgres/postgres/blob/REL_17_4/COPYRIGHT)). This shim is available for ARM and x86, and also comes with a pure C implementation. 

Portions of this code are extracted from a Postgres patch by dgrowleyml(at)gmail(dot)com's code at [pgsql-hackers](https://www.postgresql.org/message-id/CAApHDvpLXwMZvbCKcdGfU9XQjGCDm7tFpRdTXuB9PVgpNUYfEQ@mail.gmail.com).

As a result of these changes I see a 55% speedup on Apple Silicon M1 for a string set of benchmarks. 

```
== Encoding strings (2524333 bytes)
        json (local)    62.000 i/100ms
        json (local)    662.248 (± 4.5%) i/s    (1.51 ms/i) -      3.348k in   5.065760s

Normalize to 393501 byte
== Encoding strings (2524333 bytes)
       json (2.10.2)    43.000 i/100ms
       json (2.10.2)    425.189 (± 6.4%) i/s    (2.35 ms/i) -      2.150k in   5.079077s
```

The `dumps` benchmark, which covers the JSON files in `benchmark/data/*.json` – with the exception of `canada.json` – , reported a speedup of ~13% on Apple M1.

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -414,6 +414,222 @@ static void convert_UTF8_to_ASCII_only_JSON(search_state *search, const unsigned
     }
 }
 
+/* Converts input string (in ptr and len) to a JSON string, (without the wrapping
+ * '"' characters) in FBuffer buffer. ASCII control characters (0x00-0x1F),
+ * dquote, and backslash are escaped, but no other characters.
+ *
+ * This implementation is not suited for ascii_only and script_safe mode.
+ */
+#include "./simd.h"
+
+#define SIMD_BATCH_SIZE sizeof(Vector8)
+
+#define SIMD_MINIMAL_SIZE 8
+
+static inline bool needs_json_escaping(const char* ptr) {
+    Vector8 chunk;
+
+    vector8_load(&chunk, (const uint8 *)ptr);
+
+    /* Break for ASCII control characters (0x00-0x1F), dquote, and backslash. */
+    return vector8_has(chunk, '"') || vector8_has(chunk, '\\') || vector8_has_le(chunk, 0x1F);
+}
+
+
+static const char universal_escape_table[256] = {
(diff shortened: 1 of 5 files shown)


radiospiel:
Unfortunately I only saw https://github.com/ruby/json/pull/743 once the work here was finished. Also, initially I worked against 2.9.1, and I think the changes in that area should prepare for #743, but they are cannot be easily reused for my code.

radiospiel:
@byroot here is again another license to take a look.

radiospiel:
Note: see [this](https://github.com/ruby/json/pull/768#issuecomment-2728622581) for discussing licensing concerns.

byroot:
Aside from the licensing concern, SIMD use requires runtime checking of the CPU capabilities, we shouldn't assume the CPU has NEON or SSE2 (I know most do, but still).

radiospiel:
I can pick this up if we could address the licensing concerns.

byroot:
I think the licensing is OK, I need to read the PG license a second time, but it does indeed seem very BSD like, hence should allow sub-licensing. 

In other word, we can vendor it, with the license header, and keep distributing the `json` gem without changing the license.

byroot:
That said, out of your 3 PRs, I'd suggest to prioritize the float one. While float aren't super common, that's were `ruby/json` is the most behind.

byroot:
Superseeded by https://github.com/ruby/json/pull/743

PR #775, opened 2025-03-26
Remove `unparse` and `restore` aliases.

casperisfine:
These were deprecated 16 years ago.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,10 @@
 * Deprecate `JSON.load_default_options`.
 * Deprecate `JSON.unsafe_load_default_options`.
 * Deprecate `JSON.dump_default_options`.
+* Remove deprecated `JSON.restore` method.
+* Remove deprecated `JSON.unparse` method.
+* Remove deprecated `JSON.fast_unparse` method.
+* Remove deprecated `JSON.pretty_unparse` method.
 
 ### 2025-03-12 (2.10.2)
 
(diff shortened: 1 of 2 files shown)


Earlopain:
Hm, should this have been done in a patch version? From the comment the intention was to deprecate, but it hasn't actually happened (only nodoc). It breaks `sidekiq-batch` because it uses `JSON.unparse` [here](https://github.com/breamware/sidekiq-batch/blob/1f3d547a5f446c057046b57f7549066ed9b87379/lib/sidekiq/batch.rb#L50).

A fix is simple enough but there was no deprecation warning for this alias that would have helped me out here.

casperisfine:
> but it hasn't actually happened (only nodoc).

`nodoc` means it isn't public API hence can be removed, so it was effectively deprecated. I get that an actual deprecation warning would have been better.

That being said, since I never saw `unparse` used in the past I assumed it was just some left over, a quick code search shows there is indeed some use in the wild: https://github.com/search?q=JSON.unparse+language%3ARuby&type=code&l=Ruby. 

So perhaps it should go through explicit deprecation.

Earlopain:
I would say `nodoc` implies that is was documented at some point (although I haven't been around Ruby 16 years ago, so I wouldn't know personally). Looks like for `sidekiq-batch`, it was only added 8 years ago with https://github.com/breamware/sidekiq-batch/commit/77920626357daee34d87a07b3ba7f421747149e2. Can't really tell why that was done though :shrug:

PR #782, opened 2025-04-07
Suppress -Wunknown-pragmas warnings

nobu:
https://github.com/ruby/json/actions/runs/14124416241/job/39570345675#step:5:19


byroot (approved):

byroot:
I'll probably refactor that code further so these are no longer needed. But happy to merge this as a quick fix.

PR #787, opened 2025-04-24
Bring back the numerous deprecated alias

byroot:
This time with explicit deprecation warnings.

FYI: @earlopain


Earlopain:
Thanks (:

PR #792, opened 2025-04-27
Handle non-string keys returning immediate values via `to_s`

byroot:
We can't directly call `RBASIC_CLASS` as the return value of `to_s` may be an immediate.

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -789,6 +789,21 @@ struct hash_foreach_arg {
     int iter;
 };
 
+static VALUE
+convert_string_subclass(VALUE key)
+{
+    VALUE key_to_s = rb_funcall(key, i_to_s, 0);
+
+    if (RB_UNLIKELY(rb_type(key_to_s) != T_STRING)) {
+        VALUE cname = rb_obj_class(key);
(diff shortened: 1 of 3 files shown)


ext/json/ext/generator/generator.c
```diff
@@ -789,6 +789,21 @@ struct hash_foreach_arg {
     int iter;
 };
 
+static VALUE
+convert_string_subclass(VALUE key)
+{
+    VALUE key_to_s = rb_funcall(key, i_to_s, 0);
+
+    if (RB_UNLIKELY(rb_type(key_to_s) != T_STRING)) {
```
eregon, line 797: `RB_TYPE_P()` might be faster: https://github.com/ruby/ruby/blob/3ec7bfff2e7ac4f6f69d26676edcfd2e73ea3b05/include/ruby/internal/value_type.h#L376

eregon: Ah it doesn't handle T_STRING explicitly currently, so I guess it's the same, but maybe in the future it might?

byroot: This is an unlucky path, so I don't think it matter too much, but it's nicer to read, so I changed it.

PR #800, opened 2025-05-10
Favor decimal notation over scientific notation for floats

byroot:
e.g.
```
JSON.dump(1746861937.7842371)
```

master:
```
"1.7468619377842371e+9"
```

This branch and older json versions:
```
1746861937.7842371
```

In the end it's shorter, and according to `canada.json` benchmark performance is the same.

@radiospiel any objections?

For reference, sidekiq was using a bad regexp to parse JSON and was broken by this change: https://github.com/sidekiq/sidekiq/issues/6700. I'm not considering using scientific notation to workaround that bug, but I think it's better overall.

### DIFF
--- a/ext/json/ext/vendor/fpconv.c
+++ b/ext/json/ext/vendor/fpconv.c
@@ -340,7 +340,7 @@ static int emit_digits(char* digits, int ndigits, char* dest, int K, bool neg)
     }
 
     /* write decimal w/o scientific notation */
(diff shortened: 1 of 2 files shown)


radiospiel:
This looks good to me, thanks for taking care of it.

byroot:
Interestingly, JRuby always has been using scientific notation for this.

PR #803, opened 2025-05-16
fix for pretty_generate throwing wrong number of arguments error

CodyBriscoe:
Found this error to first appear in 2.11.0 as a breaking change in one of our codebases.

When `State === opts` is false, pretty_generate would eventually call generate with 3 arguments:
`State.generate(obj, options, nil)`

However if `State === opts`, it would originally try to call `state.generate(obj)` with only one argument. However, `state` just returns the State class itself `JSON::Ext::Generator::State` and expects 3 arguments.

This change aligns the behavior with the regular [JSON.generate](https://github.com/ruby/json/blob/3226913a577a0543bb59be426e47033879835b57/lib/json/common.rb#L424) method to call `opts.generate` instead.


byroot:
Oh, that's a dumb mistake. Could you add a regression test?

I can release the fix quickly.

CodyBriscoe:
Sure thing, will have that up in a moment

CodyBriscoe:
@byroot let me know if you want anything tweaked or added

PR #806, opened 2025-05-17
A few minor consistency changes

GrantBirki:
This pull requests does a few minor tweaks (no behavioral changes) to a few things I noticed when just passing over some files. For example, two files I looked at had a mix of `JSON.parser` and `::JSON.parser` so I made them consistent. Same thing with `JSON::<method>` and `JSON.method`. I just made them consistent in the files I looked at.

If this just seems like a nitpick, feel free to close the PR - I won't be offended.

### DIFF
--- a/lib/json/common.rb
+++ b/lib/json/common.rb
@@ -1072,7 +1072,7 @@ module ::Kernel
     end
 
     objs.each do |obj|
-      puts JSON::generate(obj, :allow_nan => true, :max_nesting => false)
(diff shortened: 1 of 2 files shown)


lib/json/generic_object.rb
```diff
@@ -18,7 +18,7 @@ def json_creatable?
 
       def json_create(data)
         data = data.dup
-        data.delete JSON.create_id
+        data.delete ::JSON.create_id
```
byroot, line 21: If anything we should do the opposite, the `::` is superfluous.

PR #811, opened 2025-05-31
Speed up String encoding for non-SSE and NEON CPUs

radiospiel:
This PR tests for the presence of characters to be escaped using SIMD with non-hardware-supported instructions. We are testing 8 chars in one round, following a code excerpt from https://lemire.me/blog/2025/04/13/detect-control-characters-quotes-and-backslashes-efficiently-using-swar/

For strings that don't require escapes I have seen speedups of around 15% with this change.

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -7,6 +7,95 @@
 
 #include "simd.h"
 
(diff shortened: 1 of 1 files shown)


radiospiel:
sry, pushed the wrong branch :(

PR #812, opened 2025-05-31
Speed up String encoding on CPUs without SIMD hardware support

radiospiel:
This change tests for the presence of characters to be escaped on CPUs without special SIMD instructions. We are testing 8 chars in one round, following a code excerpt from https://lemire.me/blog/2025/04/13/detect-control-characters-quotes-and-backslashes-efficiently-using-swar/

For strings that don't require escapes I have seen speedups of around 15% with this change.

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -162,8 +162,25 @@ static const unsigned char escape_table_basic[256] = {
 
 static unsigned char (*search_escape_basic_impl)(search_state *);
 
+inline bool has_json_escapable_byte(uint64_t x) {
+    uint64_t is_ascii = 0x8080808080808080ULL & ~x;
+    uint64_t xor2 = x ^ 0x0202020202020202ULL;
+    uint64_t lt32_or_eq34 = xor2 - 0x2121212121212121ULL;
+    uint64_t sub92 = x ^ 0x5C5C5C5C5C5C5C5CULL;
+    uint64_t eq92 = (sub92 - 0x0101010101010101ULL);
+    return ((lt32_or_eq34 | eq92) & is_ascii) != 0;
+}
+
 static inline unsigned char search_escape_basic(search_state *search)
 {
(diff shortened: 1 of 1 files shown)


ext/json/ext/generator/generator.c
```diff
@@ -162,8 +162,25 @@ static const unsigned char escape_table_basic[256] = {
 
 static unsigned char (*search_escape_basic_impl)(search_state *);
 
+inline bool has_json_escapable_byte(uint64_t x) {
```
byroot, line 165:
```suggestion
static inline bool has_json_escapable_byte(uint64_t x) {
```

ext/json/ext/generator/generator.c
```diff
@@ -162,8 +162,25 @@ static const unsigned char escape_table_basic[256] = {
...
+    uint64_t sub92 = x ^ 0x5C5C5C5C5C5C5C5CULL;
+    uint64_t eq92 = (sub92 - 0x0101010101010101ULL);
+    return ((lt32_or_eq34 | eq92) & is_ascii) != 0;
+}
+
 static inline unsigned char search_escape_basic(search_state *search)
 {
+    while (search->ptr <= search->end - 8) {
+        uint64_t* pi = (uint64_t*)(search->ptr);
+        if(has_json_escapable_byte(*pi)) {
```
byroot, line 178:
```suggestion
        uint64_t *pi = (uint64_t *)search->ptr;
        if (has_json_escapable_byte(*pi)) {
```

byroot: Not a huge deal, but rather than a function that return a boolean and fallback to the slow path, we could use the same structure as SIMD codepaths (return a mask that gives us the position of characters to escape).

byroot: That being said, I'm not sure if this PR is really worth it, as we kinda assume the overwhelming majority of users will have SIMD available. So it's questionable whether complexifying the fallback implementation bring any benefit.

PR #814, opened 2025-06-03
Forward-port Ractor change from ruby/ruby

hsbt:
from https://bugs.ruby-lang.org/issues/21262


test/json/ractor_test.rb
```diff
@@ -25,7 +25,7 @@ def test_generate
       end
       expected_json = JSON.parse('{"a":2,"b":3.141,"c":"c","d":[1,"b",3.14],"e":{"foo":"bar"},' +
                       '"g":"\\"\\u0000\\u001f","h":1000.0,"i":0.001}')
-      actual_json = r.take
+      actual_json = r.value
```
byroot, line 28: We can't just change it, as tests have to pass on older versions. I'll take care of it.

PR #816, opened 2025-06-13
Optimize 'json_parse_string' using ARM Neon.

samyron:
This PR uses ARM Neon instructions to optimize the `json_parse_string` function in the parser. 

I did refactor the `simd.h` from the generator to `ext/json/ext/simd/simd.h` and reference it from both the generator and the parser.

Similar to the generator, this can be disabled using the `--disable-parser-use-simd` flag.

## Benchmarks

Machine: Macbook Air M1

`master` branch at commit `41d89748fab7343bfd59e55bc171b97d61ab2bb8`

`neon-simd-parser` commit `e884e5025613db8bf3259286c17ecfa9f48aa537`

```
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +YJIT +PRISM [arm64-darwin24]
               after   965.000 i/100ms
               after      9.768k (± 0.7%) i/s  (102.38 μs/i) -     49.215k in   5.038831s

Comparison:
              before:     8467.9 i/s
               after:     9767.6 i/s - 1.15x  faster


== Parsing twitter.json (567916 bytes)
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +YJIT +PRISM [arm64-darwin24]
               after    88.000 i/100ms
               after    871.681 (± 1.3%) i/s    (1.15 ms/i) -      4.400k in   5.048502s

Comparison:
              before:      809.6 i/s
               after:      871.7 i/s - 1.08x  faster


ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +YJIT +PRISM [arm64-darwin24]
               after    41.000 i/100ms
               after    407.105 (± 7.1%) i/s    (2.46 ms/i) -      2.050k in   5.068731s

Comparison:
              before:      416.8 i/s
               after:      407.1 i/s - same-ish: difference falls within error


== Parsing ohai.json (32444 bytes)
ruby 3.4.2 (2025-02-15 revision d2930f8e7a) +YJIT +PRISM [arm64-darwin24]
               after   980.000 i/100ms
               after      9.622k (± 8.2%) i/s  (103.93 μs/i) -     48.020k in   5.041688s

Comparison:
              before:    10094.9 i/s
               after:     9622.1 i/s - same-ish: difference falls within error
```


ext/json/ext/parser/parser.c
```diff
@@ -857,32 +859,89 @@ static const bool string_scan[256] = {
...
+        // https://lemire.me/blog/2025/04/13/detect-control-characters-quotes-and-backslashes-efficiently-using-swar/
+        const uint8x16_t too_low_or_dbl_quote = vcltq_u8(veorq_u8(chunk, vdupq_n_u8(2)), vdupq_n_u8(33));
+
+        uint8x16_t has_backslash = vceqq_u8(chunk, vdupq_n_u8('\\'));
+        uint8x16_t needs_escape  = vorrq_u8(too_low_or_dbl_quote, has_backslash);
+
+        // Benchmarking this on an M1 Macbook Air shows that this is faster than
+        // using the neon_match_mask function (see the generator.c code) and bit
+        // operations to find the first match.
+        if (vmaxvq_u8(needs_escape)) {
```
byroot, line 899: I assume this is why you didn't share the between the two? Because I'd much prefer share it for maintainability reasons.

ext/json/ext/simd/simd.h
```diff
@@ -59,6 +59,20 @@ static SIMD_Implementation find_simd_implementation(void) {
 #define HAVE_SIMD 1
 #define HAVE_SIMD_NEON 1
 
+#if (defined(__GNUC__ ) || defined(__clang__))
+#define FORCE_INLINE __attribute__((always_inline))
+#else
+#define FORCE_INLINE
+#endif
```
samyron, line 66: This macro is now defined in 3 files. If we move this above the `#ifdef JSON_ENABLE_SIMD` so it's always available we can remove this from the `generator/generator.c` and `parser/parser.c`.

samyron:
[SSE2 Support](https://github.com/samyron/json/pull/5). I can merge that to include it in this PR or I can create a separate PR after this is merged if you prefer. The benchmarks look pretty good.

samyron:
> What I'd like to figure out, is a way to share the search function, because the logic is exactly the same, and the code non-trivial.
> 
> But that's tricky because the implementation in `generator.c` is currently tied to `FBuffer`.

I think I might know a way to do that, though I'm not entirely sure what the API will be. However, the idea is to create some sort of `vector_iterator` structure. Something like the following:

```
typedef struct _vector_iterator {
  // Does this iterator have another vector's worth of data?
  int (*has_chunk)(struct _vector_iterator *);

  // Return a pointer to the start of the first chunk.
  char * (*ptr)(struct _vector_iterator *);

  // Advance by an entire vector. We may not need this if we have an vector_size function.
  void (*advance)(struct _vector_iterator *);

  // Advance the underlying data structure by some number of bytes.
  void (*advance_by)(struct _vector_iterator *, int);
} vector_iterator;
```

Then we can refactor the search code to use the iterator.

```
static inline FORCE_INLINE bool string_scan_vector(vector_iterator *iter)
{
    while (iter->has_chunk(iter)) {
        char *ptr = iter->ptr(iter);

        // The actual NEON code.
        uint64_t mask = neon_step(ptr);

        if (mask) {
            iter->advance_by(iter, trailing_zeros64(mask) >> 2);
            return true;
        }

        iter->advance(iter);
    }
    
    // Will need to handle the remainder somewhere else. Or make the vector_iterator handle scalar, byte-by-byte code.

    return 0;
}
```

In the parser there might be an implementation something like:

```
typedef struct _parser_vector_iterator {
   vector_iterator *iter;
   JSON_ParserState *state;
   int vector_size;
} parser_vector_iterator;

static int has_chunk(vector_iterator *iter) {
   parser_vector_iterator *i = (parser_vector_iterator *)iter;
   return i->state->cursor <= i->state->end;
}

static char *ptr(vector_iterator *iter) {
   parser_vector_iterator *i = (parser_vector_iterator *)iter;
   return i->state->cursor;
}

void advance_by(struct _vector_iterator *iter, int num_bytes) {
   parser_vector_iterator *i = (parser_vector_iterator *)iter;
   i->state->cursor += num_bytes;
}

void advance(struct _vector_iterator *iter) {
   parser_vector_iterator *i = (parser_vector_iterator *)iter;
   i->state->cursor += i->vector_size;
}

// To create this we'd have something like

static void somewhere(JSON_ParserState *state) {
  parser_vector_iterator iter = {
    .ptr = ptr,
    .has_chunk = has_chunk,
    .advance_to = advance_to,
    .advance = advance,
    .state = state,
    .vector_size = sizeof(uint8x16_t) // Assuming Neon  
  }

  // Use this iterator somewhere... 
}
```

As I type this I'm not entirely sure that's the correct level of abstraction, but hopefully it gets the idea across. We can hopefully have the same interface but the implementations within the Parser and Generator can depend on their own local data structures.

froydnj:
> With the assumption that x86-64 chips may not have SSE2

All x86-64 chips have SSE2; it comes as part of the base ISA.

samyron:
The SIMD code should now be shared by the parser and generator.

ext/json/ext/simd/simd.h
```diff
@@ -0,0 +1,205 @@
...
+}
+
+typedef struct _neon_string_scan_iterator {
+  int (*has_next_vector)(void *, size_t);
+  const char *(*ptr)(void *);
+  void (*advance_by)(void *, size_t);
+  void (*set_match_mask)(void *, uint64_t);
+} NeonStringScanIterator;
+
+static inline FORCE_INLINE uint64_t compute_chunk_mask_neon(const char *ptr)
```
samyron, line 84: I'm not set on this function name.

byroot:
I'll try to find some time to attempt a solution without function pointers to see if it fare any better.

samyron:
I should note that on `clang 17` and `gcc 14` on macOS and `clang 16`, `clang 17`, `clang 18` and `gcc 13.3` on Ubuntu inline all of the SIMD and related `iterator` code into `json_parse_string`. A quick check of the generator code on Ubuntu using `gcc 13.3` also shows the code is inlined.

samyron:
Real world parsing benchmarks on x86-64 using the SSE2 code on a `Intel(R) Core(TM) i7-8850H` using `gcc 13.3`.


```
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after   548.000 i/100ms
               after      5.415k (± 4.0%) i/s  (184.67 μs/i) -     27.400k in   5.068215s

Comparison:
              before:     4135.9 i/s
               after:     5415.0 i/s - 1.31x  faster


== Parsing twitter.json (567916 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after    42.000 i/100ms
               after    425.069 (± 5.2%) i/s    (2.35 ms/i) -      2.142k in   5.052769s

Comparison:
              before:      395.0 i/s
               after:      425.1 i/s - same-ish: difference falls within error


ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after    18.000 i/100ms
               after    195.139 (± 6.1%) i/s    (5.12 ms/i) -    972.000 in   5.003306s

Comparison:
              before:      183.7 i/s
               after:      195.1 i/s - same-ish: difference falls within error


== Parsing ohai.json (32444 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after   497.000 i/100ms
               after      4.985k (± 3.3%) i/s  (200.59 μs/i) -     25.347k in   5.089864s

Comparison:
              before:     4486.4 i/s
               after:     4985.3 i/s - 1.11x  faster


Run #2

ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after   548.000 i/100ms
               after      5.288k (± 4.7%) i/s  (189.11 μs/i) -     26.852k in   5.089488s

Comparison:
              before:     4208.4 i/s
               after:     5287.9 i/s - 1.26x  faster


== Parsing twitter.json (567916 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after    40.000 i/100ms
               after    442.922 (± 3.8%) i/s    (2.26 ms/i) -      2.240k in   5.065280s

Comparison:
              before:      389.7 i/s
               after:      442.9 i/s - 1.14x  faster


ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after    17.000 i/100ms
               after    189.731 (± 3.7%) i/s    (5.27 ms/i) -    952.000 in   5.025374s

Comparison:
              before:      191.2 i/s
               after:      189.7 i/s - same-ish: difference falls within error


== Parsing ohai.json (32444 bytes)
ruby 3.4.1 (2024-12-25 revision 48d4efcb85) +PRISM [x86_64-linux]
               after   447.000 i/100ms
               after      4.511k (± 2.4%) i/s  (221.67 μs/i) -     22.797k in   5.056402s

Comparison:
              before:     4082.0 i/s
               after:     4511.3 i/s - same-ish: difference falls within error
```

samyron:
There are no more function pointers / iterator structure. The code now uses pointers as output variables to modify the necessary state and pass the match `mask` back to the caller.

PR #818, opened 2025-06-23
Deprecate duplicate keys in object

byroot:
There are few legitimate use cases for duplicate keys, and can in some case be exploited.

Rather to always silently accept them, we should emit a warning, and in the future require to explictly allow them.

TODO: jruby version


miharekar:
Is there a way to improve the debug message somewhat? It's impossible to find what's causing these in my test runs 😅 

```
...
/Users/miharekar/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/json-2.13.1/lib/json/common.rb:338: warning: detected duplicate keys in JSON object. This will raise an error in json 3.0 unless enabled via `allow_duplicate_key: true` at line 1 column 1632
/Users/miharekar/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/json-2.13.1/lib/json/common.rb:338: warning: detected duplicate keys in JSON object. This will raise an error in json 3.0 unless enabled via `allow_duplicate_key: true` at line 1 column 10
/Users/miharekar/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/json-2.13.1/lib/json/common.rb:338: warning: detected duplicate keys in JSON object. This will raise an error in json 3.0 unless enabled via `allow_duplicate_key: true` at line 1 column 1402
/Users/miharekar/.local/share/mise/installs/ruby/3.4.5/lib/ruby/gems/3.4.0/gems/json-2.13.1/lib/json/common.rb:338: warning: detected duplicate keys in JSON object. This will raise an error in json 3.0 unless enabled via `allow_duplicate_key: true` at line 1 column 1632
...
```

byroot:
> Is there a way to improve the debug message somewhat?

https://github.com/ruby/json/pull/831

byroot:
> It's impossible to find what's causing these in my test runs

But also you can turn the warning into an error, that should give you a backtrace.

miharekar:
> you can turn the warning into an error

I assume this is coming from one of my dependencies so not that easy to do that without knowing which `#parse` call is throwing this.

byroot:
Yeah, I'm currently working on getting the warning emitted at the caller. But you can do something like:

```ruby
JSON.singleton_class.prepend(Module.new {
  def parse(source, opts = nil)
    opts ||= {}
    opts[:allow_duplicate_key] = false
    super
  end
})
```

NB: Not saying this is fine, just giving you a quick solution until this is improved.

byroot:
With https://github.com/ruby/json/pull/832, the warning should point at the caller.

miharekar:
Awesome, #832 did the trick!

Thanks so much for this and all that you do! 🙏

renchap:
If this can be useful to someone, I noticed that most of the time this happened in our app was caused by 1 hash with symbol keys being merged with one hash with string keys, then serialised to JSON, then parsed.

PR #820, opened 2025-06-30
Suppress -Wunused-function

k0kubun:
I saw a compiler warning when compiling ruby/ruby. This PR attempts to suppress it by making it an inline function, which is usually not warned by `-Wunused-function`.

```
In file included from ../../../../ext/json/parser/parser.c:23:
../../../../ext/json/parser/../simd/simd.h:186:28: warning: ‘find_simd_implementation’ defined but not used [-Wunused-function]
  186 | static SIMD_Implementation find_simd_implementation(void) {
      |                            ^~~~~~~~~~~~~~~~~~~~~~~~
```

I'm not sure why it's happening (or `HAVE_SIMD` is not defined) on my ruby build, but it does seem unused in parser.c when `HAVE_SIMD` is not defined. So it does seem like a legit warning in some cases.

### DIFF
--- a/ext/json/ext/simd/simd.h
+++ b/ext/json/ext/simd/simd.h
@@ -59,7 +59,7 @@ static inline int trailing_zeros(int input) {
 #include <arm_neon.h>
 
 #define FIND_SIMD_IMPLEMENTATION_DEFINED 1
-static SIMD_Implementation find_simd_implementation(void) {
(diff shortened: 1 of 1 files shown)


nobu:
That warning is due to simd/conf.rb was required and evaluated only once, as described at #819.

PR #826, opened 2025-07-23
Fix missing write barrier on Generator State

jhawthorn:
Found by wbcheck. This should only be a problem when state is an object rather than stack allocated.

```
WBCHECK ERROR: Missed write barrier detected!
  Parent object: 0x7b7b8487c450 (wb_protected: true)
       rb_obj_info_dump: 0x00007b7b8487c450 JSON/Generator/State/JSON::Ext::Generator::State JSON/Generator/State
  Reference counts - snapshot: 1, writebarrier: 0, current: 6, missed: 5
  Missing reference to: 0x7b7b82f35a10
    rb_obj_info_dump: 0x00007b7b82f35a10 T_STRING/String  len: 1, capa: 15 "1"
  Missing reference to: 0x7b7b82f35e90
    rb_obj_info_dump: 0x00007b7b82f35e90 T_STRING/String  len: 1, capa: 15 "2"
  Missing reference to: 0x7b7b83629e50
    rb_obj_info_dump: 0x00007b7b83629e50 T_STRING/String  len: 1, capa: 15 "3"
  Missing reference to: 0x7b7b83b62190
    rb_obj_info_dump: 0x00007b7b83b62190 T_STRING/String  len: 1, capa: 15 "4"
  Missing reference to: 0x7b7b83629490
    rb_obj_info_dump: 0x00007b7b83629490 T_STRING/String  len: 1, capa: 15 "5"
```

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -1609,15 +1609,30 @@ static VALUE cState_buffer_initial_length_set(VALUE self, VALUE buffer_initial_l
     return Qnil;
 }
 
+struct configure_state_data {
+    JSON_Generator_State *state;
+    VALUE vstate;  // Ruby object that owns the state, or Qfalse if stack-allocated
+};
+
+static inline void state_write_value(struct configure_state_data *data, VALUE *field, VALUE value)
+{
+    if (RTEST(data->vstate)) {
+        RB_OBJ_WRITE(data->vstate, field, value);
+    } else {
+        *field = value;
+    }
+}
+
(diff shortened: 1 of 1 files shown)


byroot:
Good catch. I don't think this can realistically cause a bug, but I'm all for silencing what wbcheck finds.

byroot:
Yeah, when I say unlikely, is that the object wouldn't be old by that point (unless GC.stress etc). So defiitely real, but also definitely almost impossible to happen in real world code, hence likely why it was missed.

This is more me stating that this doesn't warrant a bug fix release on its own.

Issue #859, opened 2025-09-18
IndexOutOfBoundsException during SWAR string encoding


sl0thentr0py:
We're seeing this error on our CI with v2.14.0 on `jruby-9.4.12.0`

```
     Failure/Error: Unable to find java.base/java.nio.Buffer.checkIndex(Buffer.java to read failed line
     Java::JavaLang::IndexOutOfBoundsException:
     # java.base/java.nio.Buffer.checkIndex(Buffer.java:749)
     # java.base/java.nio.HeapByteBuffer.getInt(HeapByteBuffer.java:439)
     # json.ext.SWARBasicStringEncoder.encode(SWARBasicStringEncoder.java:47)
     # json.ext.StringEncoder.generate(StringEncoder.java:171)
     # json.ext.Generator.generateString(Generator.java:672)
     # json.ext.Generator.generateFor(Generator.java:168)
     # json.ext.Generator.processEntry(Generator.java:650)
     # json.ext.Generator.generateHash(Generator.java:580)
     # json.ext.Generator.generateFor(Generator.java:174)
     # json.ext.Generator.processEntry(Generator.java:650)
     # json.ext.Generator.generateHash(Generator.java:580)
     # json.ext.Generator$HashHandler.generate(Generator.java:489)
     # json.ext.Generator$HashHandler.generate(Generator.java:475)
     # json.ext.Generator$Handler.generateToBuffer(Generator.java:264)
     # json.ext.Generator$Handler.generateNew(Generator.java:258)
     # json.ext.Generator.generateJson(Generator.java:85)
     # json.ext.GeneratorState.generate(GeneratorState.java:236)
     # json.ext.GeneratorState.generate(GeneratorState.java:138)
```

https://github.com/getsentry/sentry-ruby/actions/runs/17797166961/job/50676908911#step:6:289

byroot:
Sadly I don't know much about Java / JRuby, but I assume it's a result of https://github.com/ruby/json/pull/835.

@headius / @samyron could one of you take care of this?

byroot:
Isn't it what https://github.com/ruby/json/blob/55552cafe2c4d54658fddd7a8c6134a42b0200b5/Rakefile#L143 is for?

Why would it suddenly be a problem? I don't want to have to deal with figuring out how to downgrade to Java 8 on my machine.

samyron:
It looks like there are a few instances in this class where I missed that...

byroot:
> Could this actually be a bug?

[...]

PR #860, opened 2025-09-18
fix issue reading off the end of the ByteBuffer if ptr > 0

samyron:
Hopefully resolves https://github.com/ruby/json/issues/859


byroot:
> Do you think you could craft a regression test?

Ok, I figured how to repro:

```ruby
require "json"
s = "01234567890"
JSON.dump(s[2..-1])
```

byroot:
I pushed a regression test.

samyron:
I added a print statement to see what `ptr` is in that example:

```
rb(main):001> require "json"
irb(main):002> s = "01234567890"
irb(main):003> JSON.dump(s[2..-1])
ptr: 2
```

PR #866, opened 2025-09-19
Small JRuby build fixes

### LINKED ISSUE #862: Modify JRuby ext build to specify JDK release version
The javac build of the JRuby extension currently only sets the source and target flags, which only guarantee the version of bytecode emitted. In order to avoid binding to newer versions of JDK APIs, We should be passing the release flag when building on JDK versions 9 or higher.

This is a secondary issue found in #859 caused by the extension using newer versions of the Buffer class's API.

headius:
Two fixes here:

* Add JRuby build output to the CLEAN FileList and Ragel-generated code to the CLOBBER FileList.
* When building on JDK 9+, use `--release` to specify Java 8 compatibility (fixes #862).


headius:
This is good to go. I did not make any changes to fix `rake clean` and `rake clobber` on CRuby.

PR #868, opened 2025-10-03
Run jruby-head on Windows

headius:
Last few failures on Windows caused by jruby/jruby#9018 were fixed in jruby/ruby#9019.


headius:
Green! This is minor so I'll merge.

PR #871, opened 2025-10-04
Fix sliced string escaping

### LINKED ISSUE #870: JRuby isn't escaping strings correctly
I've encounted a weird issue that I don't know how to debug.  This is using JRuby 10.0.2.0.  I used the version installed by `ruby-build` using `rbenv install`.

There's a rake task in ruby/net-imap that generates some code (mostly regular expressions and hashes) from RFC tables.  First it parses the RFC and writes the data into a JSON file.  Then it reads that JSON file and generates the code.  It was broken into two parts so rake can conditionally run the generator only if the JSON or the generator itself has changed.

Anyway, it's not generating the JSON correctly, and I can't figure out why.  To replicate:

From a clone of https://github.com/ruby/net-imap, run `bundle exec rake saslprep_rb`.

To dig deeper, run `irb`, and from that irb console, do the following:
```ruby
load "./rakelib/string_prep_tables_generator.rb"
generator = StringPrepTablesGenerator.new
rfc_path  = generator.rfc_filename
rfc_text  = File.read rfc_path
parsed    = generator.send(:parse_rfc_text, rfc_text)
```

This `parsed` variable now holds the data that will be generated and written to disk and then reread in a later step.  But a couple of strings toward the bottom show the issue:
```ruby
irb(main):052> d_1 = parsed["titles"]["D.1"]
=> "Characters with bidirectional property \"R\" or \"AL\""
irb(main):053> d_1 == d_1.dump.undump
=> true
irb(main):054> d_1.bytes == d_1.dump.undump.bytes
=> true
irb(main):055> d_1.encoding == d_1.dump.undump.encoding
=> true
irb(main):060> puts JSON.generate d_1.dump.undump
"Characters with bidirectional property \"R\" or \"AL\""
=> nil
irb(main):061> puts JSON.generate d_1
"Characters with bidirectional property "R" or "AL\""
=> nil
```

Same with:
```ruby
irb(main):065> d_2 = parsed["titles"]["D.2"]
=> "Characters with bidirectional property \"L\""
irb(main):067> puts JSON.generate d_2.dump.undump
"Characters with bidirectional property \"L\""
=> nil
irb(main):068> puts JSON.generate d_2
"Characters with bidirectional property "L\""
=> nil
```

What's going on with the quoting there?!  What's different about the string and its `dump.undump` counterpart?

tompng:
Fixes #870

```java
bb = ByteBuffer.wrap(array, offset, length);
bb.getLong(); // Reads from offset and increment offset. ByteBuffer's offset is mutable.
bb.getLong(idx); // Reads from idx, not from offset+idx
bb.getLong(offset + idx); // What we need
```

### `ByteBuffer.wrap(array, 0, ptr + len)` vs `ByteBuffer.wrap(array)`
`array.length` does not match `ptr+len` for a sliced string. Specifying `ptr+len` seems more safe.

### Test string
`'01234567890123456789"a"b"c"d"e"f"g"h'[20, 15]`
First 15 bytes don't need escape.
15 = 8 + 4 + 3, which can test getLong, getInt and remaining bytes escaping

### DIFF
--- a/java/src/json/ext/SWARBasicStringEncoder.java
+++ b/java/src/json/ext/SWARBasicStringEncoder.java
@@ -27,9 +27,10 @@ public class SWARBasicStringEncoder extends StringEncoder {
         // slice a string, the underlying byte array is the same, but the
         // begin index and real size are different. When reading from the ptrBytes
         // array, we need to always add ptr to the index.
-        ByteBuffer bb = ByteBuffer.wrap(ptrBytes, ptr, len);
+        ByteBuffer bb = ByteBuffer.wrap(ptrBytes, 0, ptr + len);
+
         while (pos + 8 <= len) {
-            long x = bb.getLong(pos);
+            long x = bb.getLong(ptr + pos);
             if (skipChunk(x)) {
                 pos += 8;
                 continue;
@@ -48,7 +49,7 @@ public class SWARBasicStringEncoder extends StringEncoder {
         }
 
         if (pos + 4 <= len) {
-            int x = bb.getInt(pos);
+            int x = bb.getInt(ptr + pos);
             if (skipChunk(x)) {
                 pos += 4;
             }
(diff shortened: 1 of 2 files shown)


byroot:
Thank you!

headius:
Aha, another buffer-wrapping issue. Thanks for jumping on this!

PR #887, opened 2025-11-01
parser.c: Optimize `json_parse_digits`

byroot:
We can use `ctz` builtin to get the number of consecutive digits.

```
== Parsing float parsing (2251051 bytes)
ruby 3.4.6 (2025-09-16 revision dbd83256b1) +YJIT +PRISM [arm64-darwin24]
               after    25.000 i/100ms
               after    231.963 (± 0.0%) i/s    (4.31 ms/i) -      1.175k in   5.065467s

Comparison:
              before:      215.3 i/s
               after:      232.0 i/s - 1.08x  faster
```

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -612,12 +612,14 @@ json_eat_whitespace(JSON_ParserState *state)
                 while (rest(state) > 8) {
                     uint64_t chunk;
                     memcpy(&chunk, state->cursor, sizeof(uint64_t));
-                    size_t consecutive_spaces = trailing_zeros64(chunk ^ 0x2020202020202020) / CHAR_BIT;
-
-                    state->cursor += consecutive_spaces;
-                    if (consecutive_spaces != 8) {
-                        break;
+                    if (chunk == 0x2020202020202020) {
+                        state->cursor += 8;
+                        continue;
                     }
+
+                    uint32_t consecutive_spaces = trailing_zeros64(chunk ^ 0x2020202020202020) / CHAR_BIT;
+                    state->cursor += consecutive_spaces;
+                    break;
                 }
 #endif
                 break;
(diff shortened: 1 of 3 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -1094,20 +1094,27 @@ static inline int json_parse_digits(JSON_ParserState *state, uint64_t *accumulat
         // From: https://github.com/simdjson/simdjson/blob/32b301893c13d058095a07d9868edaaa42ee07aa/include/simdjson/generic/numberparsing.h#L333
         // Branchless version of: http://0x80.pl/articles/swar-digits-validate.html
         uint64_t match = (next_8bytes & 0xF0F0F0F0F0F0F0F0) | (((next_8bytes + 0x0606060606060606) & 0xF0F0F0F0F0F0F0F0) >> 4);
+        size_t consecutive_digits = trailing_zeros64(match ^ 0x3333333333333333) / CHAR_BIT;
```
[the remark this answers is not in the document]
byroot: Yes thank you I just figured this out:

> int __builtin_clz (unsigned int x)

>    Returns the number of leading 0-bits in x, starting at the most significant bit position. If x is 0, the result is undefined. 

https://gcc.gnu.org/onlinedocs/gcc/Bit-Operation-Builtins.html

I think GCC assume this case never happens so it removes that branch.

ext/json/ext/simd/simd.h
```diff
@@ -19,7 +19,7 @@ typedef enum {
 static inline uint32_t trailing_zeros64(uint64_t input)
 {
 #if HAVE_BUILTIN_CTZLL
-    return __builtin_ctzll(input);
+    return input ? __builtin_ctzll(input) : 0;
```
byroot: cc @tompng here how I fixed it.

byroot: Nevermind 🤦

ext/json/ext/simd/simd.h
```diff
@@ -18,6 +22,10 @@ typedef enum {
 
 static inline uint32_t trailing_zeros64(uint64_t input)
 {
+#ifdef JSON_DEBUG
+    assert(input > 0); // __builtin_ctz(0) is undefined behavior
```
byroot: I'm doubting these `asserts` can actually fail. Not too sure what the compilation flags are on CI, but seems like they are optimized out even with  `JSON_DEBUG`. 

I'm unsure what's the best way to add assertions in ruby extensions. Perhaps I should just use `RUBY_ASSERT` and add a debug ruby in the CI matrix?

PR #893, opened 2025-11-12
[DOC] Doc for class String

BurdetteLamar:
Text for `String` class doc, for `String.json_create`, and for `String#to_json_raw`.

`:nodoc:` for `String#to_json_raw_object`, which I don't think users need.


byroot (commented):
Thank you for this.

However this is an API that I do intend to extract in an optional gem, because it's very dangerous an inefficient. So I'm not sure it really needs much advertisement.

PR #896, opened 2025-11-18
C Extension: Use #if instead of #ifdef for JSON_DEBUG

samyron:
While profiling the C extension, I kept hitting code within the `#ifdef JSON_DEBUG`. Using `#ifdef` checks if the identifier is defined, not the value.

I did not have the `JSON_DEBUG` environment variable set but noticed the following in `json.h`:
```
#ifdef RUBY_DEBUG
#ifndef JSON_DEBUG
#define JSON_DEBUG RUBY_DEBUG
#endif
#endif
```

Adding `#error "RUBY_DEBUG defined but not JSON_DEBUG"` after the `#ifndef` and running `rake` outputs the following:

```
In file included from ../../../../../../ext/json/ext/generator/generator.c:1:
../../../../../../ext/json/ext/generator/../json.h:82:2: error: "RUBY_DEBUG defined but not JSON_DEBUG"
   82 | #error "RUBY_DEBUG defined but not JSON_DEBUG"
      |  ^
1 error generated.
```

I should note that I am _not_ running debug build of Ruby. Adding a `printf("RUBY_DEBUG=%lu\n", RUBY_DEBUG);` outputs a lot of `RUBY_DEBUG=0`.

My assumption is we want this code to be excluded if JSON_DEBUG is 0.

### DIFF
--- a/ext/json/ext/generator/extconf.rb
+++ b/ext/json/ext/generator/extconf.rb
@@ -6,7 +6,7 @@ if RUBY_ENGINE == 'truffleruby'
 else
   append_cflags("-std=c99")
   $defs << "-DJSON_GENERATOR"
-  $defs << "-DJSON_DEBUG" if ENV["JSON_DEBUG"]
+  $defs << "-DJSON_DEBUG" if ENV.fetch("JSON_DEBUG", "0") != "0"
 
   if enable_config('generator-use-simd', default=!ENV["JSON_DISABLE_SIMD"])
     load __dir__ + "/../simd/conf.rb"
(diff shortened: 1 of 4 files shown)


ext/json/ext/generator/extconf.rb
```diff
@@ -6,7 +6,7 @@
 else
   append_cflags("-std=c99")
   $defs << "-DJSON_GENERATOR"
-  $defs << "-DJSON_DEBUG" if ENV["JSON_DEBUG"]
+  $defs << "-DJSON_DEBUG" if ENV.fetch("JSON_DEBUG", "0") != "0"
```
samyron: This is optional but it allows `JSON_DEBUG=0` to not trigger debugging code. Feel free to remove if this isn't needed.

PR #903, opened 2025-11-25
Move depth to generate_json_data

etiennebarrie:
I want to remove the memcpy of `generate_json_data` and remove the FIXME by putting the mutating depth where it should be, in `generate_json_data`. The State object itself still has a depth, for configuration.

The idea is to have a new `generate_json_data::depth` field and stop mutating the old `JSON_Generator_State::depth` one, and then at the end, copy the depth back to it. This doesn't need to happen in all cases, e.g. `generate_new` doesn't need to copy it back, and this is what allows us not to do the memcpy it in the first place.

This still keeps `generate_new` to let it have different behaviors between `JSON::Coder` (calling `dump` does not result in the `depth` of the `State` changing if a `JSON::NestingError` occurs) and `JSON::State` (which does). I called that last behavior deprecated because I have another PR to start deprecating this behavior.

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -1539,12 +1539,6 @@ static VALUE cState_generate_new(int argc, VALUE *argv, VALUE self)
 
     GET_STATE(self);
 
-    JSON_Generator_State new_state;
-    MEMCPY(&new_state, state, JSON_Generator_State, 1);
-
-    // FIXME: depth shouldn't be part of JSON_Generator_State, as that prevents it from being used concurrently.
-    new_state.depth = 0;
-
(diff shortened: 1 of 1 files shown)


ext/json/ext/generator/generator.c
```diff
@@ -972,6 +973,8 @@ static inline VALUE vstate_get(struct generate_json_data *data)
     if (RB_UNLIKELY(!data->vstate)) {
         vstate_spill(data);
     }
+    GET_STATE(data->vstate);
+    state->depth = data->depth;
```
etiennebarrie: Ouch actually this is bad, the State is mutated while it might be frozen, which also means that multiple Ractors could write to this concurrently. Fix coming up.

PR #906, opened 2025-11-28
Test and restore behavior around `to_json` changing depth

etiennebarrie:
Follow up to #903

When serializing an Array, and one of the elements of the Array requires calling `to_json`, if the depth is changed, it will be used for the next entries, which wasn't the case before
5abd43490714460e1ec1fa382f82a09485eed563, and is not the case with TruffleRuby and JRuby.

Additionally, with TruffleRuby and JRuby the state's depth after the `to_json` call is used to close the Array, which isn't the case with CRuby.

### DIFF
--- a/java/src/json/ext/Generator.java
+++ b/java/src/json/ext/Generator.java
@@ -462,10 +462,10 @@ public final class Generator {
             generateFor(context, session, element, buffer);
         }
 
-        int oldDepth = state.decreaseDepth();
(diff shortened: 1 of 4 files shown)


byroot:
> Additionally, with TruffleRuby and JRuby the state's depth after the `to_json` call is used to close the Array, which isn't the case with CRuby.

Can't we fix that behavior rather than to encode it in test?

PR #917, opened 2025-12-11
Add `allow_control_characters` parsing option

byroot:
While it's not allowed by the spec, some parsers like Oj do accept it, and it can be blocking a transition.

Having this feature can help people migrate.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 ### Unreleased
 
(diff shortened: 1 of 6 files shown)


byroot:
Actually slightly better patch:

```ruby
module JSONAllowControlCharacters
  def parse(doc, opts = nil)
    super(doc, { allow_control_characters: true }.merge(opts || {}))
  end
end
JSON.singleton_class.prepend(JSONAllowControlCharacters)
```

PR #920, opened 2025-12-31
[parser.c] Various string unescape optimizations

samyron:
This PR contains 4 commits which optimize `json_string_unescape`. Admittedly, I almost didn't create this PR as I don't love the Apple Aarch64-specific code... however I decided to leave it up to the maintainers of this library to decide if this code is worth accepting.

I omitted the SWAR / ARM Neon unless the code is being compiled on Apple platforms as a quick benchmark on my x86 laptop with the SWAR `find_backslash` and custom `json_memcpy` code showed it was no faster `master`. Looking at the `memcpy` and `memchr` on Ubuntu showed a AVX optimized libc. At the time the code was structured a bit differently and it was only copying `sizeof(uint64_t)` bytes at a time on x86. I could try implementing `find_backslash` and `json_memcpy` using SSE2 if desired.  I should note that on `memchr` on macOS _does_ seem to be optimized using ARM Neon instructions. Looking at the assembly, the implementation is similar though not quite the same. I think the performance improvement comes from reduced function call overhead / branching as `find_backslashes` ends up inlined.

With respect to keeping track of the `additional_backslashes` instead of `has_more`, if the string to unescape ends with a long run of characters that do not have any more backslashes, we can save a `memchr` and directly copy to the end of the string. 

I'm cool if you decide not to accept any of these and simply close this PR. These are just some random things I've been testing out.

# Benchmarks

These were run on an Macbook Air M1.

## With only the `addtional_backslashes` commit:

The benchmark:
```
benchmark_parsing 'a bunch of escapes followed by a long string without escapes', JSON.generate("some\nescapes\twith\"a long\"string after\fso\twe\ncan test\nno\ffinal\tmemchr" * 4 + "this is some longish string! " * 512)
```

```
== Parsing a bunch of escapes followed by a long string without escapes (15162 bytes)
               after    33.105k i/100ms
               after    345.312k (± 2.3%) i/s    (2.90 μs/i) -      1.755M in   5.083936s

Comparison:
              before:   307715.4 i/s
               after:   345311.9 i/s - 1.12x  faster
```

## `additional_backslashes` and `find_backslashes`:

```
== Parsing a bunch of escapes followed by a long string without escapes (15162 bytes)
               after    32.551k i/100ms
               after    351.976k (± 2.6%) i/s    (2.84 μs/i) -      1.790M in   5.089949s

Comparison:
              before:   307223.8 i/s
               after:   351976.5 i/s - 1.15x  faster


               after     1.094k i/100ms
               after     10.824k (± 0.8%) i/s   (92.39 μs/i) -     54.700k in   5.054133s

Comparison:
              before:    10415.9 i/s
               after:    10823.5 i/s - 1.04x  faster
```

## `additional_backslashes`, `find_backslashes`, `json_memcpy`:

```
== Parsing a bunch of escapes followed by a long string without escapes (15162 bytes)
               after    32.384k i/100ms
               after    342.436k (± 3.2%) i/s    (2.92 μs/i) -      1.716M in   5.017785s

Comparison:
              before:   288084.1 i/s
               after:   342436.2 i/s - 1.19x  faster


               after     1.128k i/100ms
               after     11.272k (± 0.7%) i/s   (88.71 μs/i) -     56.400k in   5.003607s

Comparison:
              before:    10586.3 i/s
               after:    11272.4 i/s - 1.06x  faster
```

### `additional_backslashes`, `find_backslashes`, `json_memcpy`, `unescape_unicode`:

```
               after     1.037k i/100ms
               after     11.368k (± 0.6%) i/s   (87.96 μs/i) -     57.035k in   5.017163s

Comparison:
              before:    10600.4 i/s
               after:    11368.4 i/s - 1.07x  faster
```

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -479,21 +479,16 @@ static const signed char digit_values[256] = {
 
 static uint32_t unescape_unicode(JSON_ParserState *state, const unsigned char *p)
 {
-    signed char b;
-    uint32_t result = 0;
-    b = digit_values[p[0]];
-    if (b < 0) raise_parse_error_at("incomplete unicode character escape sequence at %s", state, (char *)p - 2);
-    result = (result << 4) | (unsigned char)b;
-    b = digit_values[p[1]];
-    if (b < 0) raise_parse_error_at("incomplete unicode character escape sequence at %s", state, (char *)p - 2);
-    result = (result << 4) | (unsigned char)b;
-    b = digit_values[p[2]];
-    if (b < 0) raise_parse_error_at("incomplete unicode character escape sequence at %s", state, (char *)p - 2);
-    result = (result << 4) | (unsigned char)b;
-    b = digit_values[p[3]];
-    if (b < 0) raise_parse_error_at("incomplete unicode character escape sequence at %s", state, (char *)p - 2);
-    result = (result << 4) | (unsigned char)b;
-    return result;
+    signed char b0 = digit_values[p[0]];
+    signed char b1 = digit_values[p[1]];
+    signed char b2 = digit_values[p[2]];
+    signed char b3 = digit_values[p[3]];
+    
+    if ((b0 | b1 | b2 | b3) < 0) {
+        raise_parse_error_at("incomplete unicode character escape sequence at %s", state, (char *)p - 2);
+    }
+    
+    return ((uint32_t)b0 << 12) | ((uint32_t)b1 << 8) | ((uint32_t)b2 << 4) | (uint32_t)b3;
 }
 
 #define GET_PARSER_CONFIG                          \


ext/json/ext/parser/parser.c
```diff
@@ -479,21 +479,16 @@ static const signed char digit_values[256] = {
...
-    if (b < 0) raise_parse_error_at("incomplete unicode character escape sequence at %s", state, (char *)p - 2);
-    result = (result << 4) | (unsigned char)b;
-    b = digit_values[p[3]];
-    if (b < 0) raise_parse_error_at("incomplete unicode character escape sequence at %s", state, (char *)p - 2);
-    result = (result << 4) | (unsigned char)b;
-    return result;
+    signed char b0 = digit_values[p[0]];
+    signed char b1 = digit_values[p[1]];
+    signed char b2 = digit_values[p[2]];
+    signed char b3 = digit_values[p[3]];
```
byroot: I think there is a risk of reading past the end of the buffer here.

Prior `digit_values[p[0]] < 0` would return if `p[0]` is a NULL byte.

I think this optimization does make sense, but we need to ensure we can actually read 4 bytes.

byroot:
> as I don't love the Apple Aarch64-specific code... however I decided to leave it up to the maintainers of this library to decide if this code is worth accepting.

To be frank with you, my opinion on the matter is that Apple hardware is overwhelmingly used in development, almost never in production, so it's not really worth taking on extra complexity just for that environment, unless of course the gain is massive.

I do however like your two optimizations that aren't specific to macOS+aarch64.

byroot:
Alright, I cherry picked d21d9362fa36bd1b34c53e28b04ba6e91115a3be and 976ba36629b2d20fa4de3da12ed405ab178deeba on master.

Thank you for the PR!

Issue #939, opened 2026-02-19
Allow option to ignore illegal escaped characters


gokuu:
Hello!

I realize that what I'm asking is probably a long shot, but it doesn't hurt to try.

In a nutshell, v2.17.0 of the `json` ruby gem has broke our systems. We ingest a lot of JSON files tht contain customer-produced data. This means that we do not control the JSON that is parsed.
Before v2.17.0, everything worked without a hitch. However, after upgrading `json` due to it being a dependency of a gem we use, all hell broke loose, and several (over 50% for some of our major customers) JSON files stopped being ingested. 
After investigating the reason, we realized it was due to having escaped characters that now fail the parsing process.

So, I was wondering if it would be possible to add an extra option to the [`JSON_ParserConfig` struct](https://github.com/ruby/json/blob/master/ext/json/ext/parser/parser.c#L330-L341) that would allow us to keep the behavior that existed prior to v2.17.0?

byroot:
Please see the discussion at https://github.com/ruby/json/pull/917#issuecomment-3914683549

In your case the best way forward is just to apply that monkey patch.

PR #940, opened 2026-02-19
Add `allow_invalid_escape` parsing option

byroot:
Fix: https://github.com/ruby/json/issues/939


headius:
@byroot Thanks for handling the JRuby extension as well!

Issue #944, opened 2026-02-27
dump bigdecimals as floats


HoneyryderChuck:
been working on migrating an application from oj to json gem, and I believe this is the last hurdle to cross.

Oj dumps bigdecimals as floats, whereas json dumps them as strings:

```ruby
bigd = BigDecimal("1.12345678912345678912345")
Oj.dump(bigd) #=> "1.12345678912345678912345"
JSON.dump(bigd) #=> "\"1.12345678912345678912345\""
```

I understand the reasons for the default behaviour, however some clients support full precision floats (even if over the standard 32bit) and start shipping strings would be the compatibility breaking one could avoid.

Is it possible to support this via an additional `dump_default_options`?

FWIW `JSON.load` already supports parsing those via `decimal_class: decimal_class` load option.

PR #946, opened 2026-03-05
Add automatic conversion to BigDecimal when using `JSON.parse` with high-precision floats

HoneyryderChuck:
Automatically parses JSON numbers with >17 significant digits as BigDecimal, falls back to using Float otherwise. This behaviour is conditional on "bigdecimal" being around, and will fall back on current behaviour of converting to Float with precision loss otherwise.

This behaviour matches what `Oj.load` does [by default](https://github.com/ohler55/oj/blob/develop/lib/oj/mimic.rb#L19-L20), and is a proposal designed to smoothen migrations away from `oj`. 

@byroot while I started #944 mostly focused on encoding bigdecimals, which we agreed it could be done via some monkeypatching which eventually can live in activesupport, the main blocker for our internal migration has actually been parsing. I mentioned that, by using `JSON.load` with `:decimal_class` option, but in practice this is turning out not to be true, for a few reasons.

The main one is that `JSON.load` and `JSON.parse` have some surprisingly different behaviour. Take these examples:

```ruby
# JSON.load is liberal with empty string
JSON.load("") #=> nil
JSON.parse("") #=> unexpected end of input at line 1 column 1 (JSON::ParserError)

# JSON.load ignores inputs which can't be converted
JSON.load([]) #=> nil
JSON.parse([]) #=> no implicit conversion of Array into String (TypeError)
# sidenote, but the above is made even more confusing with oj's monkey-patching of JSON.load and JSON.parse...
JSON.load([]) #=> ArgumentError: parser io argument must be a String or respond to readpartial() or read(). (ArgumentError)
JSON.parse([]) #=> []
```

The difference in behaviour and the (in our case) widespread use of `JSON.parse` (partly driven by a rubocop cop which prevents usage of `JSON.load` for security reasons) make a rewrite to `JSON.load` quite risky. 

Another drawback is that 3rd party libs mostly use `JSON.parse` under the hood, and few/none allow passing down your own json decoder, so one could convert everything to `JSON.load` potentially and still trip on it down the line.

A way to use `JSON.parse` while supporting this conversion to bigdecimals would be ideal. `json` does not support a similar `:bigdecimal_load` option, and `JSON.parse` does not support global options anyway (not that I'm advocating for it), hence why I went with this "no config" approach based on having bigdecimals loaded. 

There's also the argument of backwards compatibility (a bigdecimal is not a float) which could surprise someone in case of a release. I'm not sure what's the best way to deal with this, but FWIW I still consider that the benefit (no precision loss) would outweight the backwards incompatible change cost. 

This change was originally authored by @samgiddins while trying to solve the same issue internally. FYI I removed everything encoding-related, removed the block always trying to load bigdecimal, and tweaked the check for const defined.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -5,7 +5,7 @@
 static VALUE mJSON, eNestingError, Encoding_UTF_8;
 static VALUE CNaN, CInfinity, CMinusInfinity;
 
-static ID i_new, i_try_convert, i_uminus, i_encode;
+static ID i_new, i_try_convert, i_uminus, i_encode, i_BigDecimal;
 
 static VALUE sym_max_nesting, sym_allow_nan, sym_allow_trailing_comma, sym_allow_control_characters,
              sym_allow_invalid_escape, sym_symbolize_names, sym_freeze, sym_decimal_class, sym_on_load,
@@ -839,9 +839,22 @@ static inline VALUE json_decode_float(JSON_ParserConfig *config, uint64_t mantis
         return rb_funcallv(config->decimal_class, config->decimal_method_id, 1, &text);
     }
 
+    // Automatically use BigDecimal for high-precision decimals (> 17 significant digits)
+    // that would lose precision when converted to Float
+    // We check the mantissa value itself rather than the digit count to handle
+    // numbers like 0.000000022471348024634545 which have leading zeros
+    if (RB_UNLIKELY(mantissa_digits > 17 && mantissa >= 100000000000000000ULL)) {
+        if (rb_const_defined(rb_cObject, i_BigDecimal)) {
+            VALUE text = rb_str_new(start, end - start);
+            return rb_funcallv(rb_mKernel, i_BigDecimal, 1, &text);
+        }
+        // If BigDecimal is not available, fall back to regular float parsing
+        return json_decode_large_float(start, end - start);
+    }
+
     // Fall back to rb_cstr_to_dbl for potential subnormals (rare edge case)
     // Ryu has rounding issues with subnormals around 1e-310 (< 2.225e-308)
-    if (RB_UNLIKELY(mantissa_digits > 17 || mantissa_digits + exponent < -307)) {
+    if (RB_UNLIKELY(mantissa_digits + exponent < -307)) {
         return json_decode_large_float(start, end - start);
     }
 
@@ -1669,6 +1682,7 @@ void Init_parser(void)
     i_try_convert = rb_intern("try_convert");
     i_uminus = rb_intern("-@");
     i_encode = rb_intern("encode");
+    i_BigDecimal = rb_intern("BigDecimal");
 
     binary_encindex = rb_ascii8bit_encindex();
     utf8_encindex = rb_utf8_encindex();
--- a/test/json/json_addition_test.rb
+++ b/test/json/json_addition_test.rb
@@ -3,7 +3,7 @@ require_relative 'test_helper'
 require 'json/add/core'
 require 'json/add/complex'
 require 'json/add/rational'
-require 'json/add/bigdecimal'
+# require 'json/add/bigdecimal' - removed in favor of simpler BigDecimal handling
 require 'json/add/ostruct'
 require 'json/add/set'
 require 'date'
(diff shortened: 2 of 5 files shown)


HoneyryderChuck:
@byroot just updated my snippet with the following 2 changes:

* We're now using `BigDecimal#n_significant_digits` instead of `BigDecimal#precision`, as the latter didn't take into account leading zeros.
* I posted a second solution where the significant digits count was ported to ruby. This avoids the extra bigdecimal allocation of the first solution (in the hot case where it converts to a float), at the cost of string byte walking, a more involved process than what the proposed C patch does (but YJIT can optimize fairly well). 

Our benchmarks also revealed that, regardless of which solution you go with, the bulk of the performance overhead is caused just by setting `:decimal_class`; when that happens, an optimized branch is chosen where the byte subarray will be converted into a ruby string (the optimized path converts it straight into a float), then incur the cost of "C to ruby" call to invoke the `try_convert` function. Here's a snippet of it:

```ruby
require 'benchmark/ips'
require 'json'
require 'bigdecimal'

module BigDecimalSignificantDigits
  # logic above...
end

module RubySignificantDigits
  # logic above ...
end


def benchmark_parsing(name, json_output)
  puts "== Parsing #{name} (#{json_output.size} bytes)"

  Benchmark.ips do |x|
    x.report('json') { JSON.parse(json_output) }
    x.report('Float') { JSON.parse(json_output, decimal_class: Float) }
    x.report('BigDecimal') { JSON.parse(json_output, decimal_class: BigDecimal) }
    x.report('BigDecimal#n_significant_digits') { JSON.parse(json_output, decimal_class: BigDecimalSignificantDigits) }
    x.report('ruby significant_digits') { JSON.parse(json_output, decimal_class: RubySignificantDigits) }

    x.compare!(order: :baseline)
  end
  puts
end

benchmark_parsing 'small nested array', JSON.dump([[1, 2, 3, 4, 5]] * 10)
benchmark_parsing 'small hash', JSON.dump({ 'username' => 'jhawthorn', 'id' => 123, 'event' => 'wrote json serializer' })
benchmark_parsing 'test from oj', <<~JSON
  {"a":"Alpha","b":true,"c":12345,"d":[true,[false,[-123456789,null],3.9676,["Something else.",false],null]],
  "e":{"zero":null,"one":1,"two":2,"three":[3],"four":[0,1,2,3,4]},"f":null,
  "h":{"a":{"b":{"c":{"d":{"e":{"f":{"g":null}}}}}}},"i":[[[[[[[null]]]]]]]}
JSON
```

sharing the results in my machine (actually, docker container, so run your own if necessary):

```
== Parsing small nested array (121 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [aarch64-linux]
# ...
Comparison:
                json:  1263493.7 i/s
          BigDecimal:  1183296.9 i/s - same-ish: difference falls within error
               Float:  1179992.5 i/s - same-ish: difference falls within error
BigDecimal#n_significant_digits:  1150776.8 i/s - same-ish: difference falls within error
ruby significant_digits:  1138455.2 i/s - same-ish: difference falls within error

== Parsing small hash (65 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [aarch64-linux]
# ...
Comparison:
                json:  2897382.3 i/s
BigDecimal#n_significant_digits:  2446507.2 i/s - 1.18x  slower
          BigDecimal:  2426340.8 i/s - 1.19x  slower
               Float:  2377540.5 i/s - 1.22x  slower
ruby significant_digits:  2181776.1 i/s - 1.33x  slower

== Parsing test from oj (258 bytes)
ruby 3.4.8 (2025-12-17 revision 995b59f666) +YJIT +PRISM [aarch64-linux]
# ...
Comparison:
                json:   456729.8 i/s
ruby significant_digits:   408883.8 i/s - same-ish: difference falls within error
               Float:   392566.8 i/s - same-ish: difference falls within error
          BigDecimal:   361843.0 i/s - 1.26x  slower
BigDecimal#n_significant_digits:   338971.7 i/s - 1.35x  slower
```

I didn't post the results with bigger documents, but it gets worse depending on the size of payload (a test against a >2MB json file showed performance degraded 11x or worse for any solution declaring any `:decimal_class`).

For now we're not too concerned about the impact of it (don't think any of our business flows requires parsing that large of a JSON document, but considering the work you did to make json perform on par, and now even better than oj, and this performance degradation gap that may affect only <1% of cases but still affect them regardless, can I ask you if you'd reconsider the proposed patch, perhaps reframed as a new feature, for the coming v3 you're working on? Definitely not a global option (I know you want to get rid of those), but something scope to `JSON::Coder` perhaps, akin to `:bigdecimal_load`? Happy to rework this PR and add the missing jruby logic if necessary.

PR #948, opened 2026-03-05
Reject negative depth

thesmartshadow:
This PR prevents native hangs/crashes when `depth` is negative under pretty-printing.

- Reject negative `depth` with a clear `ArgumentError` in both the State setter and options application.
- ~~Add defense-in-depth overflow guards in fbuffer capacity growth and indentation repeat paths.~~
- Add a regression test ensuring negative depth fails fast instead of hanging/crashing.

All tests pass: `bundle exec rake test`.


thesmartshadow:
thanks @byroot

PR #971, opened 2026-04-15
Fix handling out of of range exponent in numbers

### LINKED ISSUE #970: `JSON.parse` returns incorrect finite values for very large decimal exponents
## Summary

`JSON.parse` returns incorrect finite values for decimal numbers with very large exponents. The result appears to behave as if the exponent is narrowed to a 32-bit signed integer before float conversion, so exponents near `2**32` wrap around instead of overflowing or underflowing consistently with the normal Ruby float parser.

This affects both the JSON version bundled with Ruby 4.0.2 and the latest JSON gem release I could install from RubyGems.

## Tested Versions

Default JSON bundled with Ruby 4.0.2:

```text
ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin24]
json 2.19.2
affected
```

Latest JSON gem installed from RubyGems:

```text
gem install json --no-document
Successfully installed json-2.19.3

ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin24]
json 2.19.3
affected
```

## Reproducer

```ruby
require "json"

puts RUBY_DESCRIPTION
puts "json #{JSON::VERSION}"

failed = false

[
  "1e4294967296",
  "1e-4294967296",
  "99999999999999999e-4294967296",
  "1e4294967295",
  "1e4294967297",
].each do |number|
  actual = JSON.parse(number)
  expected = Float(number)

  if actual == expected
    puts "ok: #{number} => #{actual.inspect}"
  else
    failed = true
    puts "mismatch: #{number}"
    puts "  expected: #{expected.inspect}"
    puts "  actual:   #{actual.inspect}"
  end
end

exit(false) if failed
```

Observed output with `json` 2.19.3:

```text
ruby 4.0.2 (2026-03-17 revision d3da9fec82) +PRISM [arm64-darwin24]
json 2.19.3
mismatch: 1e4294967296
  expected: Infinity
  actual:   1.0
mismatch: 1e-4294967296
  expected: 0.0
  actual:   1.0
mismatch: 99999999999999999e-4294967296
  expected: 0.0
  actual:   1.0e+17
mismatch: 1e4294967295
  expected: Infinity
  actual:   0.1
mismatch: 1e4294967297
  expected: Infinity
  actual:   10.0
```

## Expected Behavior

The parser should not wrap the exponent and return a finite value with a different mathematical meaning from the JSON numeric literal.

For consistency with existing `JSON.parse` behavior on large but non-wrapping exponents, these values should either follow the normal Ruby float conversion result or raise an error if the JSON parser does not want to produce non-finite Ruby floats. For example, on the same runtime:

```ruby
JSON.parse("1e10000")  # => Infinity
JSON.parse("1e-1000")  # => 0.0
```

## Root Cause Notes

In `ext/json/parser/parser.c`, exponent parsing appears to accumulate exponent digits into a `uint64_t` and then narrow the accumulated value to `int32_t` before calling the Ryu float conversion path:

```c
uint64_t abs_exponent = 0;
int exponent_digits = json_parse_digits(state, &abs_exponent);
exponent = negative_exponent ? -((int32_t)abs_exponent) : ((int32_t)abs_exponent);
```

For exponent `4294967296`, that cast produces zero on Ruby 4.0.2 `arm64-darwin24`. The fast float path then converts the mantissa with an exponent near zero instead of falling back to the original string or treating the exponent as overflowed.

byroot:
Fix: https://github.com/ruby/json/issues/970

If the parsed exponent overflows a `int32_t` passing it to ryu is incorrect.

We could pass it to `rb_cstr_to_dbl` but then Ruby will emit an annoying warning, instead we can coerce to `0.0` and `Inf`.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -855,12 +855,20 @@ NOINLINE(static) VALUE json_decode_large_float(const char *start, long len)
 /* Ruby JSON optimized float decoder using vendored Ryu algorithm
  * Accepts pre-extracted mantissa and exponent from first-pass validation
  */
-static inline VALUE json_decode_float(JSON_ParserConfig *config, uint64_t mantissa, int mantissa_digits, int32_t exponent, bool negative,
+static inline VALUE json_decode_float(JSON_ParserConfig *config, uint64_t mantissa, int mantissa_digits, int64_t exponent, bool negative,
                                           const char *start, const char *end)
 {
     if (RB_UNLIKELY(config->decimal_class)) {
         VALUE text = rb_str_new(start, end - start);
         return rb_funcallv(config->decimal_class, config->decimal_method_id, 1, &text);
+    }
+
+    if (RB_UNLIKELY(exponent > INT32_MAX)) {
+        return CInfinity;
+    }
+
+    if (RB_UNLIKELY(exponent < INT32_MIN)) {
+        return rb_float_new(0.0);
     }
 
     // Fall back to rb_cstr_to_dbl for potential subnormals (rare edge case)
@@ -869,7 +877,7 @@ static inline VALUE json_decode_float(JSON_ParserConfig *config, uint64_t mantis
         return json_decode_large_float(start, end - start);
     }
 
-    return DBL2NUM(ryu_s2d_from_parts(mantissa, mantissa_digits, exponent, negative));
+    return DBL2NUM(ryu_s2d_from_parts(mantissa, mantissa_digits, (int32_t)exponent, negative));
 }
 
 static inline VALUE json_decode_array(JSON_ParserState *state, JSON_ParserConfig *config, long count)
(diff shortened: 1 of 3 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -855,21 +855,29 @@ NOINLINE(static) VALUE json_decode_large_float(const char *start, long len)
...
+static inline VALUE json_decode_float(JSON_ParserConfig *config, uint64_t mantissa, int mantissa_digits, int64_t exponent, bool negative,
                                           const char *start, const char *end)
 {
     if (RB_UNLIKELY(config->decimal_class)) {
         VALUE text = rb_str_new(start, end - start);
         return rb_funcallv(config->decimal_class, config->decimal_method_id, 1, &text);
     }
 
+    if (RB_UNLIKELY(exponent > INT32_MAX)) {
+        return CInfinity;
```
lopopolo, line 867: don't we have to coerce to a signed infinity depending on the `negative` value?

byroot: Good catch.

byroot: Fixed in 1072482184de233f55c45f875b38766b67b84f3b

PR #992, opened 2026-06-05
Fix memsize function for embedded types

byroot:
We shouldn't report the size of the embedded struct.


nobu:
"Embeddable" doesn't mean "embedded" always, I guess.

byroot:
@nobu this is handled by the TypedDate API: https://github.com/ruby/ruby/blob/2dd77a0dcd1f634d7787e62d1d735147fd73d99a/gc.c#L1224-L1226

nobu:
Thanks!

PR #994, opened 2026-06-05
JSON::ResumableParser

### LINKED ISSUE #983: Add support for parsing chunked data
This is similar to #854 but a different feature request. Both of #854 and this are for migrating to the json gem from yajl-ruby (in [Fluentd](https://github.com/fluent/fluentd). #854's description doesn't mention it but it's mentioned in a comment https://github.com/ruby/json/issues/854#issuecomment-3279115294 ) but this proposes different API. We can't migrate from yajl-ruby with API proposed in #854.

#854 proposed `parser.parse(io)` like API but what we need is `io.each_chunk {|chunk| parser.feed(chunk)}` like API. An application wants to read data not the json gem read data. `parser.parse(io)` like API reads data by the json gem. Fluentd can't migrate to yajl-ruby with the API.

Fluentd that is a event based log collector reads data from multiple sockets efficiently by IO multiplexing. We may not be able to read a whole JSON (e.g. `["a", 2, true]`) at a time with the IO multiplexing. We may get a whole JSON as multiple chunks (e.g. `["a"`, `, 2, ` and `true]`). yajl-ruby can parse chunks without concatenating all of them:

yajl-ruby:

```ruby
require "yajl"
parser = Yajl::Parser.new
parser.on_parse_complete = lambda {|x| pp x}
parser << "[\"a\""
parser << ", 2, "
parser << "true]" # lambda {|x| pp x} is called with ["a", 2, true]
```

The json gem:

```ruby
require "json"
JSON.parse(["[\"a\"", ", 2, ", "true]"].join)
```

If an application can extract a whole JSON from an input without parsing input (*), we don't need yajl-ruby style API. But an input may have multiple JSONs (like Fluentd's use case), we need yajl-ruby style API. Because we need to parse JSON to split data in an input to JSONs for this case.

(*) For example:
1. Only one JSON exists in an input
2. Each JSON is delimited by new line (JSON lines)
3. ...

## API

`JSON::Parser.new` requires a source. So it may be better that we define a new class such as `JSON::StreamParser` instead of reusing existing `JSON::Parser`.

`JSON::StreamParser` accepts a chunk not a whole JSON:

```ruby
parser = JSON::StreamParser.new
parser.parse_chunk("[\"a\"")
parser.parse_chunk(", 2, ")
parser.parse_chunk("true]")
```

Other candidate for `parse_chunk`: `consume`

We need to define callback style API to pass parsed data to application. Because one `parse_chunk` call may parse 0 or more JSONs. For example, if chunk is `[1][2]`, `[1]` and `[2]` are parsed at once.

`JSON::Parser` accepts `on_load` callback. We may resuse the name for `JSON::StreamParser` but `on_load` is for JSON value not JSON. It may be better that we use different name. For example:

```ruby
parser = JSON::StreamParser.new
parser.on_load = lambda {|x| pp x}
parser.parse_chunk("[\"a\"")
parser.parse_chunk(", 2, ")
parser.parse_chunk("true]") # lambda {x| pp x} is called with ["a", 2, true]
parser.parse_chunk("[1][2]") # lambda {x| pp x} is called twice with [1] and [2]
```

byroot:
Fix: https://github.com/ruby/json/issues/983

Numerous known issues TODO:

  - [x] Performance: This new feature shouldn't significantly degrade classic `JSON.parse`. Right now `twitter.json` is `7-10%` slower, that's not OK. We might need to duplicate the parsing loop if necessary.
    - Edit: after removing write barriers in `rvalue_stack_push` brough most of the performance back.
    - [x] Figure out a clean way not WB protect `cResumableParser` but not the embedded `rvalue_stack`.
  - [x] `object_start_cursor` recorded in frame becomes invalid if the buffer string is reallocated or spilled.
  - [x] The buffer need to be shrunk sometimes.
  - [x] Lot more testing needed.
  - [x] Unclear what to do with top level numbers.
  - [x] API is all but final
    - I'd like to be able to "pop" the value, so we don't uselessly keep a reference on it.
    - Then methods need to be documented.
  - [x] It would worth trying to make `json_parse_any` exception free.
    - Right now EOF errors have been eliminated in favor of returning `false`.
    - We could try to do the same with syntax errors.
    - But then we need to `rb_protect` when calling back into Ruby or other unsafe APIs, so perhaps it's best to just accept it.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -5,7 +5,7 @@
 static VALUE mJSON, eNestingError, eParserError, Encoding_UTF_8;
 static VALUE CNaN, CInfinity, CMinusInfinity;
 
-static ID i_new, i_try_convert, i_uminus, i_encode, i_at_line, i_at_column;
+static ID i_new, i_try_convert, i_uminus, i_encode, i_at_line, i_at_column, i_at_eos;
 
 static VALUE sym_max_nesting, sym_allow_nan, sym_allow_trailing_comma, sym_allow_comments,
              sym_allow_control_characters, sym_allow_invalid_escape, sym_symbolize_names,
@@ -57,6 +57,20 @@ typedef struct rvalue_cache_struct {
     int length;
     VALUE entries[JSON_RVALUE_CACHE_CAPA];
 } rvalue_cache;
+
+static void rvalue_cache_mark(rvalue_cache *cache)
+{
+    for (int index = 0; index < cache->length; index++) {
+        rb_gc_mark_movable(cache->entries[index]);
+    }
+}
+
+static void rvalue_cache_compact(rvalue_cache *cache)
+{
+    for (int index = 0; index < cache->length; index++) {
+        cache->entries[index] = rb_gc_location(cache->entries[index]);
+    }
+}
 
 static rb_encoding *enc_utf8;
 
@@ -206,7 +220,7 @@ static rvalue_stack *rvalue_stack_spill(rvalue_stack *old_stack, VALUE *handle, 
 
 static rvalue_stack *rvalue_stack_grow(rvalue_stack *stack, VALUE *handle, rvalue_stack **stack_ref)
 {
-    long required = stack->capa * 2;
+    long required = stack->capa ? stack->capa * 2 : RVALUE_STACK_INITIAL_CAPA;
 
     if (stack->type == RVALUE_STACK_STACK_ALLOCATED) {
         stack = rvalue_stack_spill(stack, handle, stack_ref);
@@ -222,8 +236,10 @@ static VALUE rvalue_stack_push(rvalue_stack *stack, VALUE value, VALUE *handle, 
     if (RB_UNLIKELY(stack->head >= stack->capa)) {
         stack = rvalue_stack_grow(stack, handle, stack_ref);
     }
+
     stack->ptr[stack->head] = value;
     stack->head++;
+
     return value;
 }
 
@@ -372,7 +388,7 @@ typedef struct json_frame_struct {
     enum json_frame_type type;
     enum json_frame_phase phase;
     long value_stack_head;    // rvalue_stack->head when this container opened
-    const char *start_cursor; // object frames only (the '{'); NULL otherwise
+    size_t start_offset; // object frames only (the '{'); NULL otherwise
 } json_frame;
 
 typedef struct json_frame_stack_struct {
@@ -421,7 +437,7 @@ static json_frame_stack *json_frame_stack_spill(json_frame_stack *old_stack, VAL
 
 static json_frame_stack *json_frame_stack_grow(json_frame_stack *stack, VALUE *handle, json_frame_stack **stack_ref)
 {
-    long required = stack->capa * 2;
+    long required = stack->capa ? stack->capa * 2 : JSON_FRAME_STACK_INITIAL_CAPA;
 
     if (stack->type == RVALUE_STACK_STACK_ALLOCATED) {
         stack = json_frame_stack_spill(stack, handle, stack_ref);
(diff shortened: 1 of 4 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -222,8 +222,14 @@ static VALUE rvalue_stack_push(rvalue_stack *stack, VALUE value, VALUE *handle,
     if (RB_UNLIKELY(stack->head >= stack->capa)) {
         stack = rvalue_stack_grow(stack, handle, stack_ref);
     }
+
     stack->ptr[stack->head] = value;
     stack->head++;
+
+    if (*handle) {
+        RB_OBJ_WRITTEN(*handle, Qundef, value);
+    }
```
byroot: This seem to account for most of the perf regression on `twitter.json`.

I think instead of making `rvalue_stack` WB protected, we could just not embed it, or just have a secondary non-protected object just to mark it.

kou:
Great!

Here are some random notes:

MessagePack like API (splitting a parsing API to appending new data, parsing buffer and getting parsed data) is good.

MessagePack also uses `feed` but `feed` (`a parser feeds data`) may be a bit strange.

`parser.consume(data)`, `parser.refill(data)` or something may be better. I used "consume" when I implemented a resumable parser in Apache Arrow ( https://github.com/apache/arrow/pull/6804 ) but it uses callback style API and "consume" fills a buffer and parses the buffer. With this API, "consume" may be a bit strange too.

How about `#value` returns parsing data even when `#parse` returns `false` something like the following?

```diff
diff --git a/ext/json/ext/parser/parser.c b/ext/json/ext/parser/parser.c
index 749d594..f0731ac 100644
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -2254,9 +2254,8 @@ static VALUE cResumableParser_parse(VALUE self)
 static VALUE cResumableParser_value(VALUE self)
 {
     JSON_ResumableParser *parser = cResumableParser_get(self);
-    json_frame *frame = json_frame_stack_peek(&parser->frames);
 
-    if (frame->phase == JSON_PHASE_DONE) {
+    if (parser->state.value_stack->head > 0) {
         return *rvalue_stack_peek(parser->state.value_stack, 1);
     } else {
         rb_raise(rb_eArgError, "no ready value"); // TODO: Figure out the best exception and message
```

@tompng shared an use case of resumable parser:
* It may be useful to process generative AI API response
* It's returned as a stream
* An application wants to display the response before a response isn't completed

For example:

1. Response: `{"message": "This is a response.` (not completed)
2. App: Show `This is a response`
3. Response: ` More messages.` (not completed)
4. App: Append ` More messages.` to its view
5. ... 

The above diff doesn't satisfy this use case but we can use it for simpler case such as `[1,`. We can get `[1]` before we have rest data (something like `, 2]`).

If we also provide an API that returns not processed data something like the following, we may be able to cover the generative AI API response use case: 

```diff
diff --git a/ext/json/ext/parser/parser.c b/ext/json/ext/parser/parser.c
index 749d594..d39ba11 100644
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -2263,6 +2263,15 @@ static VALUE cResumableParser_value(VALUE self)
     }
 }
 
+static VALUE cResumableParser_rest(VALUE self)
+{
+    JSON_ResumableParser *parser = cResumableParser_get(self);
+
+    return rb_str_substr(parser->buffer,
+                         parser->state.cursor - parser->state.start,
+                         parser->state.end - parser->state.cursor);
+}
+
 void Init_parser(void)
 {
 #ifdef HAVE_RB_EXT_RACTOR_SAFE
@@ -2289,6 +2298,7 @@ void Init_parser(void)
     rb_define_method(cResumableParser, "feed", cResumableParser_feed, 1);
     rb_define_method(cResumableParser, "parse", cResumableParser_parse, 0);
     rb_define_method(cResumableParser, "value", cResumableParser_value, 0);
+    rb_define_method(cResumableParser, "rest", cResumableParser_rest, 0);
 
     CNaN = rb_const_get(mJSON, rb_intern("NaN"));
     rb_gc_register_mark_object(CNaN);
```

byroot:
> but `feed` (`a parser feeds data`) may be a bit strange.

Agreed. As I was working on this, I was thinking of simply renaming it to `<<`.

> How about #value returns parsing data even when #parse returns false something like the following?

So your snippet wouldn't work, because the Hash and Array are only built once complete.

e.g. `[1, 2, 3`, the first element on the `value_stack` is `1`, not `[1, 2, 3]`

However we could indeed build the partial object on demand. It's an interesting feature, I'll see about adding it, but I think it should either be a different method or an optional paramter, e.g. `parser.value(partial: true)`, to avoid confusion.

> an API that returns not processed data something like the following,

Interesting, wouldn't be hard indeed.

byroot:
> However we could indeed build the partial object on demand.

One worry here is that while building the partial object wouldn't be too hard, it would potentially contain objects that will later be returned by `.value`. So if they are mutated it could get weird.

Alternatively we'd need to `deep_dup` but that's tricky because there is `object_class` and configurations that allow for arbitrary objects, so the type of the already parsed objects is somewhat opaque to the parser.

kou:
> > but `feed` (`a parser feeds data`) may be a bit strange.
> 
> Agreed. As I was working on this, I was thinking of simply renaming it to `<<`.

+1

> However we could indeed build the partial object on demand. It's an interesting feature, I'll see about adding it, but I think it should either be a different method or an optional parameter, e.g. `parser.value(partial: true)`, to avoid confusion.

+1
 
> One worry here is that while building the partial object wouldn't be too hard, it would potentially contain objects that will later be returned by `.value`. So if they are mutated it could get weird.
>
> Alternatively we'd need to `deep_dup` but that's tricky because there is `object_class` and configurations that allow for arbitrary objects, so the type of the already parsed objects is somewhat opaque to the parser.

I think that we can just document that "you must not change the partial object".

We already have similar cases. For example, `a = [1, 2, 3]; a.each {|x| p x; a.delete(1)}` outputs `1` and `3` not `1`, `2` and `3`.

kou:
@Watson1978 Could you try this API whether Fluentd can replace yajl-ruby dependency with json by this API or not?

Watson1978:
> @Watson1978 Could you try this API whether Fluentd can replace yajl-ruby dependency with json by this API or not?

Thanks. I will try to replace.

byroot:
I just pushed another commit that adds `JSON::ResumableParser#partial_value`, feedback still welcome.

PR #995, opened 2026-06-05
[parser.c] simplify json_value_completed transition

samyron:
This PR simplifies the transition of of the next phase of the parser in `json_value_completed`. The `enum json_frame_type` and `enum json_frame_phase` are now equal when the `frame->type` is `JSON_FRAME_ROOT`, `JSON_FRAME_ARRAY` or `JSON_FRAME_OBJECT` so the next phase equal to the current `frame->type`.

Looking at the disassembly this saves 39 instructions. 

Admittedly this may be one notch too far on the cleverness scale and an actual lookup table would be easier to reason about.


byroot:
Nah, I like it. But you'll have to rebase a bit.

PR #996, opened 2026-06-05
parser.c: Extract json_match_keyword

byroot:
Extracted from: https://github.com/ruby/json/pull/994 (trying the make it smaller so the perf regression is easier to spot)

Modern compilers shouldn't have problem computing `strlen` at compile time and generating the same code.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -1425,9 +1425,7 @@ static inline VALUE json_parse_positive_number(JSON_ParserState *state, JSON_Par
 
 static inline VALUE json_parse_negative_number(JSON_ParserState *state, JSON_ParserConfig *config)
 {
-    const char *start = state->cursor;
-    state->cursor++;
-    return json_parse_number(state, config, true, start);
+    return json_parse_number(state, config, true, state->cursor - 1);
 }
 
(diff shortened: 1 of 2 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -1452,6 +1450,22 @@ static inline void json_value_completed(json_frame *frame)
     frame->phase = (enum json_frame_phase) frame->type;
 }
 
+static inline bool json_match_keyword(JSON_ParserState *state, const char *keyword, size_t offset)
+{
+    // It is assumed that since `keyword` is always a literal, the compiler is able to constantize this
+    // `strlen` and several other computations in that routine, such as eliminating the `if (resumable)` branch.
```
eregon, line 1456: An ALWAYS_INLINE (I think available from CRuby headers) would help guarantee that, vs depending on the compiler inlining heuristics.
(`inline` doesn't mean to inline but just tolerate multiple definitions of it)

byroot: yeah we have a shim for it.

eregon: https://github.com/ruby/json/pull/998 if it's helpful

PR #998, opened 2026-06-05
ALWAYS_INLINE for json_match_keyword()

eregon:
https://github.com/ruby/json/pull/996#discussion_r3365374461


byroot:
From what I could see it's definitely being inlined already, but yeah, doesn't cost much to ensure it.

PR #989, opened 2026-06-01
Iterative parsing

kddnewton:
Switch the recursive descent to an iterative algorithm by keeping a stack of "frames". Each frame represents parsing a container (the points at which you would recurse). Within frames there are "phases"; each phase maps to effectively a token that its looking for next.

This does not get all of the way to streaming parsing (that would require partial-token support), but it gets most of the way there, and is all required pre-work.

On my machine there's no noticeable slow-down. (There's also no noticeable speed-up.) The only really noticeable side-effect is that if you pass `max_nesting: false`, this will not crash anymore from running out of stack space.

Issue #697, opened 2024-11-06
Encoding::InvalidByteSequenceError with json 2.8.1


jterapin:
Hi again! 

When using json 2.8.1 with Ruby 3.2.0 (for additional context, we use CodeBuild instances)  - been running into the following error:
```
Encoding::InvalidByteSequenceError: "\xE2" on US-ASCII (Encoding::InvalidByteSequenceError)
/codebuild/output/src3972890485/src/.gems/ruby/3.2.0/gems/json-2.8.1/lib/json/common.rb:204:in `encode'
/codebuild/output/src3972890485/src/.gems/ruby/3.2.0/gems/json-2.8.1/lib/json/common.rb:204:in `parse'
/codebuild/output/src3972890485/src/.gems/ruby/3.2.0/gems/json-2.8.1/lib/json/common.rb:204:in `parse'
/codebuild/output/src3972890485/src/.gems/ruby/3.2.0/gems/json-2.8.1/lib/json/common.rb:710:in `load'
/codebuild/output/src3972890485/src/aws-sdk-ruby/build_tools/services.rb:93:in `load_docs'
```

The `load_docs` referred above is here: https://github.com/aws/aws-sdk-ruby/blob/version-3/build_tools/services.rb#L93
Could any of the recent changes have effected this? Reverting back to json 2.7.6 has fixed this for us.

`InvalidByteSequenceError: "\xE2" on US-ASCII` feels like its trying to read it as ASCII instead of UTF-8. 

Any guidance will be appreciated. Thanks much!!!

byroot:
Yes, I'm making the library progressively stricter about encoding, this one I think I should just note in the CHANGELOG.

> InvalidByteSequenceError: "\xE2" on US-ASCII 

This means you are passing an UTF-8 strings, but it's encoded as ASCII, looking at your code, it comes directly from `File.read`, so somehow you must have `Encoding.default_external == Encoding::ASCII`.

```bash
ruby -e 'Encoding.default_external = Encoding::ASCII; p File.read("/etc/passwd").encoding'
#<Encoding:US-ASCII>
```

I think you should check your `$LANG` environment variable. But yes, either way you are passing an invalid string to JSON so it's expected it raises. We only have a small exception for binary strings but that's it.

byroot:
A quick workaround can also be:

```ruby
JSON.load(File.read(model_path('docs-2.json', models_dir), encoding: Encoding::UTF_8))
```

jterapin:
[...]

PR #1004, opened 2026-06-15
[Java Extension] Ported the C extension parser to the Java and remove ragel generated parser.

samyron:
## Overview

This PR is a port of the current C extension parser to Java. It removes the ragel generated parser.

## Implementation Notes

1. This uses the same frame-based parse loop as the C-based parser. However, since java doesn't support goto, it uses a while/switch dispatching mechanism.
1. Building a JSON Object uses `RubyHash#fastASet` instead of `RubyHash#op_aset` as strings or symbols used as keys are explicitly frozen in this module. This contributed a significant performance boost to this Parser. The keys are frozen in `cachedKey` / `internedKey` which is called via `parseString` when `isName=true`. 
1. The Parser includes a SWAR and Vector API StringScanner. The Vector API is loaded at runtime if running on a supported JDK, the module is available and the `ruby.json.useVectorizedParser=true` JVM property is set. If the Vector API is not available or explicitly disabled the SWAR implementation is used. 
1. This also uses the same `rvalue_cache` style cache as a quick cache for object keys. However, since the cache is heap allocated in Java, the size is 128 entries.
1. The SWAR consecutive digits decoder was ported over to this Java parser.
1. It retains the Java extension Unicode character validation but moves the logic from the generated ragel parser to the StringDecoder.

## Performance

These benchmarks were run on an M1 Macbook Air using jruby 10.0.5.0 and OpenJDK 64-Bit Server VM 24.0.1+9-30.

## With the SWAR StringScanner

```
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after   656.000 i/100ms
               after      6.719k (± 1.2%) i/s  (148.83 μs/i) -     34.112k in   5.076849s

Comparison:
before:     1351.3 i/s
 after:     6719.1 i/s - 4.97x  faster


== Parsing twitter.json (567916 bytes)
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after    69.000 i/100ms
               after    706.750 (± 1.1%) i/s    (1.41 ms/i) -      3.588k in   5.076758s

Comparison:
before:      113.1 i/s
 after:      706.8 i/s - 6.25x  faster


jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after    38.000 i/100ms
               after    390.007 (± 2.3%) i/s    (2.56 ms/i) -      1.976k in   5.066570s

Comparison:
before:       37.6 i/s
 after:      390.0 i/s - 10.39x  faster


== Parsing github_events.json (65130 bytes)
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after   744.000 i/100ms
               after      7.261k (± 5.0%) i/s  (137.72 μs/i) -     36.456k in   5.020636s

Comparison:
before:     1160.6 i/s
 after:     7261.2 i/s - 6.26x  faster


== Parsing semanticscholar-corpus.json (8493528 bytes)
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after     4.000 i/100ms
               after     53.332 (± 3.8%) i/s   (18.75 ms/i) -    268.000 in   5.025161s

Comparison:
before:        8.1 i/s
 after:       53.3 i/s - 6.62x  faster
```

## With the Vector API based StringScanner
```
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after   671.000 i/100ms
               after      6.880k (± 1.3%) i/s  (145.34 μs/i) -     34.892k in   5.071152s

Comparison:
before:     1361.7 i/s
 after:     6880.5 i/s - 5.05x  faster


== Parsing twitter.json (567916 bytes)
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after    72.000 i/100ms
               after    736.863 (± 1.4%) i/s    (1.36 ms/i) -      3.744k in   5.080998s

Comparison:
before:      109.3 i/s
 after:      736.9 i/s - 6.74x  faster


jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after    38.000 i/100ms
               after    394.387 (± 1.5%) i/s    (2.54 ms/i) -      1.976k in   5.010313s

Comparison:
before:       38.8 i/s
 after:      394.4 i/s - 10.17x  faster


== Parsing github_events.json (65130 bytes)
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after   777.000 i/100ms
               after      7.681k (± 2.9%) i/s  (130.20 μs/i) -     38.850k in   5.058152s

Comparison:
before:     1097.3 i/s
 after:     7680.7 i/s - 7.00x  faster


== Parsing semanticscholar-corpus.json (8493528 bytes)
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after     5.000 i/100ms
               after     56.260 (± 3.6%) i/s   (17.77 ms/i) -    285.000 in   5.065794s

Comparison:
before:        7.3 i/s
 after:       56.3 i/s - 7.67x  faster
```


samyron:
Tagging @headius for a review on this PR.

samyron:
Additional thought: We should probably add additional JRuby versions to CI and include the Vector API system properties to ensure the `VectorizedStringScanner` and `VectorizedStringEncoder` (from a previous PR) are thoroughly tested.

byroot:
Is this the parser mentioned by @enebo in https://github.com/ruby/json/issues/983#issuecomment-4603423404 or is it a concurrent effort?

samyron:
> Is this the parser mentioned by @enebo in https://github.com/ruby/json/issues/983#issuecomment-4603423404 or is it a concurrent effort?

This is a concurrent effort. I missed that comment as I didn't refer back to that issue once https://github.com/ruby/json/pull/989 was opened.

samyron:
If we are willing to accept invalid UTF-8 bytes / broken strings(https://github.com/ruby/json/issues/138) in the Java parser, this parser can be even faster:

These benchmarks compare the current commit on this PR with some work-in-progress code I have locally. Some of this new code could be applied to the commit on this branch which may help a bit too if we want to keep the broken string validation. 

```
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after   850.000 i/100ms
               after      8.785k (± 1.1%) i/s  (113.83 μs/i) -     44.200k in   5.031459s

Comparison:
before:     6572.8 i/s
 after:     8784.7 i/s - 1.34x  faster


== Parsing twitter.json (567916 bytes)
jruby 10.0.5.0 (3.4.5) 2026-04-06 5db1ba72f3 OpenJDK 64-Bit Server VM 24.0.1+9-30 on 24.0.1+9-30 +indy +jit [arm64-darwin]
               after    78.000 i/100ms
               after    777.467 (± 1.4%) i/s    (1.29 ms/i) -      3.900k in   5.016292s

Comparison:
before:      716.2 i/s
 after:      777.5 i/s - 1.09x  faster
```

This would allow removing the JRuby guard around [this method](https://github.com/ruby/json/blob/master/test/json/json_parser_test.rb#L336-L351).

headius:
This looks like great work! I will review today. It will be great to eliminate the ragel parser and finally get performance where it should be!

byroot:
> If we are willing to accept invalid UTF-8 bytes / broken strings(#138) in the Java parser,

I honestly don't really know what to do with #138. Logically we shouldn't validate the parsed strings encoding, but I suspect a lot of people rely on this, so it probably would need to be an option like `allow_duplicate_key` etc.

So it might actually make sense to let the Java version parse invalid encoding like the C version, simply for compatibility.

java/src/json/ext/Parser.java
```diff
@@ -0,0 +1,954 @@
...
+                    case '\r':
+                        cursor++;
+                        break;
+                    case '\n':
+                        cursor++;
+                        // Same heuristic from the C parser: a newline in
+                        // pretty-printed JSON is almost always followed by a run
+                        // of indentation spaces, so skip them eight at a time.
+                        while (cursor + 8 <= end) {
+                            long x = chunks.getLong(cursor);
```
headius, line 842: This `getLong` and similar calls in `StringScanner` could potentially be replaced with `VarHandle`, which allows reading a `byte[]` as `long.` However `VarHandle` is only available on Java 9+ which would make this code unusable on JRuby 9.4  on Java 8 (technically EOL but still in wide use). Not sure if we are ready to make a hard break in `json` yet.

java/src/json/ext/Parser.java
```diff
@@ -0,0 +1,954 @@
...
+                ? ((RubySymbol) entry).getBytes()
+                : ((RubyString) entry).getByteList();
+            int elen = eb.realSize();
+            if (len != elen) {
+                return len - elen;
+            }
+            byte[] ebuf = eb.getUnsafeBytes();
+            int ebeg = eb.begin();
+            for (int i = 0; i < len; i++) {
+                int cmp = (buf[off + i] & 0xFF) - (ebuf[ebeg + i] & 0xFF);
```
headius: Masking with `0xFF` can be replaced with the less arcane `Byte.toUnsignedInt` or `Byte.toUnsignedLong`.

java/src/json/ext/Parser.java
```diff
@@ -0,0 +1,954 @@
...
+                        cursor++;
+                        break;
+                    case '\n':
+                        cursor++;
+                        // Same heuristic from the C parser: a newline in
+                        // pretty-printed JSON is almost always followed by a run
+                        // of indentation spaces, so skip them eight at a time.
+                        while (cursor + 8 <= end) {
+                            long x = chunks.getLong(cursor);
+                            if (x == 0x2020202020202020L) {
```
headius: This and other important literal values should probably be in `static final long` constants.

headius (approved):
This is good work and I approve merging it. My review comments are mostly style-related but we can get more performance out of this by doing some additional profiling and reusing the parser stack and wrapper objects, using VarHandle to do long-stride reads rather than ByteBuffer, etc. That can all come after this lands, though.

headius:
> I honestly don't really know what to do with https://github.com/ruby/json/issues/138.

The strictness of the Java-based parser has not been a real barrier to anyone I know of.

We have had a few folks over the years switch to JRuby and discover they had bad string content coming in from some json source. Since we obviously never made the json parser allow such content, they all eventually fixed the bad source. 

Do we really know how bad the breakage would be if the CRuby json ext started rejecting bad UTF-8 content? I'd rather not reduce the strictness of the Java parser just for a little bit of performance.

byroot:
> Do we really know how bad the breakage would be if the CRuby json ext started rejecting bad UTF-8 content?

https://github.com/ruby/json/issues/697 comes to mind, and it was a comparatively much simpler change.

But it's really hard to tell.

As for performance, on paper I agree with you, but history have shown that many users will likely use an alternative if has the reputation to be faster, regardless of whether that really matters or not.

headius:
> many users will likely use an alternative if has the reputation to be faster

I'd like to be able to quantify that. @samyron have you already tested a version of the Java parser that is non-strict?

@byroot I assume when you suggest adding another configuration option, it would allow users to opt *out* of strictness.

Releasing a version that's strict by default would be the quickest way to find out how many people are affected, and having an opt-out config would give those users a temporary workaround. It would be nice for the Ruby json library to help eliminate bad json content rather than silently propagating it.

byroot:
> it would allow users to opt _out_ of strictness.

No in, with a warning by default.

> It would be nice for the Ruby json library to help eliminate bad json content rather than silently propagating it.

Given I'm on the receiving end of complaints, no thanks.

samyron:
> I'd like to be able to quantify that. @samyron have you already tested a version of the Java parser that is non-strict?

Yes, see the numbers in [this comment](https://github.com/ruby/json/pull/1004#issuecomment-4714444425). I have this work on a separate branch [here](https://github.com/samyron/json/tree/sm/jruby-parser-rewrite-followup).

There are two big changes on that branch:
- Adding SWAR/Vectorized scanning to the `StringDecoder#decode` to skip the non-interesting bytes.
- Removed the UTF-8 validation.

I no longer have these two changes isolated but if I recall correctly in UTF-8 heavy text the added SWAR/Vectorized overhead in the `StringDecoder#decode` was pretty significant. That's why I decided to experiment with matching the C Parser and removing that validation.

headius:
Definitely would like to get @enebo input here before merging this or combining with his parser.

enebo:
I am going to compare and contrast a bit today on this.

enebo:
Thus far, I think this version seems to be better with strings and my version is quite a bit faster for numbers (floats).  The small array and hash benches favor my branch but this version uses a pretty large starting stack which might explain that difference.  I am using raw byte[] and this uses ByteBuffer.  I think we can mix some stuff together to get best of both worlds.

The other comment is since this maps more closely structurally to the C extension then it might have an advantage over mine from a maintenance standpoint.  The parsers are pretty small overall so not a big problem but I can see if you are updating one it would be nicer for it to be closer.

I want to look a bit more but I think we can make something better than both versions as they stand.  It should be trivial to plug in my numeric parsing stuff.

enebo:
I am just dumping JSON broader test results from https://github.com/nst/JSONTestSuite.git between the two new parsers.  I think both are fine as nearly all of the failures are undefined behavior and strange enough I don't think anyone can reasonably expect any particular behavior.  This PR has one (trailing raw //)  which needs to get fixed as I think it gets stuck in a loop but that should be simple to fix.



Screenshot From 2026-06-16 10-02-35

This PR vs MRI:
Screenshot From 2026-06-16 10-01-04

My parser branch vs MRI:

Screenshot From 2026-06-16 10-02-02

byroot:
> I am just dumping JSON broader test results from https://github.com/nst/JSONTestSuite.git

Note that the Ruby harness is a bit broken: https://github.com/nst/JSONTestSuite/pull/145

But I've meant to vendor that test suite in ruby/json for a while now. I'll try to do it soon.

java/src/json/ext/Parser.java
samyron: I don't know if it would be worth it / how HotSpot would handle it but I could create an interface `DataView` or something that uses `VarHandle` if available and otherwise falls back to `ByteBuffer` if not.

The call sites would be monomorphic once the implementation is selected so I would hope that HotSpot will handle that appropriately. Something I can test later...

enebo:
This PR running on my machine using the parser benchmark 'float parsing':

```text
== Parsing float parsing (2251051 bytes)
jruby 10.1.1.0-SNAPSHOT (4.0.0) 2026-06-08 482a54b9cb OpenJDK 64-Bit Server VM 25.0.1+8-27 on 25.0.1+8-27 +indy +jit [x86_64-linux]
                json     7.000 i/100ms
                json     70.402 (± 1.4%) i/s   (14.20 ms/i) -    707.000 in  10.04508
```

but swapping in my parser's version of parseNumber:

```text
== Parsing float parsing (2251051 bytes)
jruby 10.1.1.0-SNAPSHOT (4.0.0) 2026-06-08 482a54b9cb OpenJDK 64-Bit Server VM 25.0.1+8-27 on 25.0.1+8-27 +indy +jit [x86_64-linux]
                json    16.000 i/100ms
                json    175.713 (± 1.1%) i/s    (5.69 ms/i) -      1.760k in  10.018326s
```

Now what really confuses me is here is my parser running it:

```text
== Parsing float parsing (2251051 bytes)
jruby 10.1.1.0-SNAPSHOT (4.0.0) 2026-06-08 482a54b9cb OpenJDK 64-Bit Server VM 25.0.1+8-27 on 25.0.1+8-27 +indy +jit [x86_64-linux]
                json    12.000 i/100ms
                json    124.782 (± 2.4%) i/s    (8.01 ms/i) -      1.248k in  10.006981s
```

So something magical is happening in this PRs parser which is not on mine.  Yet, most of the benchmarks on mine will edges out this one until it hits multiple byte UTF-8 characters.  Then this parser does quite a bit better again (citm_catalog and twitter).

@samyron and others watching this PR.  I am sorry that we are in the position of two complete implementations but I think we will be better for it in the end.  I am going to continue to examine the impls today and probably tomorrow.  I really like the idea of using this one because it is so close to the C version (which is good for synch'ing when new things happen).  That said, my parser is doing a little better in most of the benchmarks (twitter and citm_catalog being the exceptions --- and also probably the most real-world examples).  I think if I can get this one's speed up to mine then we should use this one.

enebo:
Parsing numbers enhancement branch is here https://github.com/enebo/json/tree/jruby-parser-rewrite-numbers

samyron:
@enebo Are the screenshots backwards? I wasn't able to reproduce the `{"a": "b"}//` results locally. I ran the full test suite against my branch and CRuby and see this:

image

enebo:
@samyron My apologies.  I did mix the two up.  I thought I was using dates for which was which and I tested my branch last week.  I clearly reversed them. The new addition of comment tests preceeding element had me have to arrange some stuff and this was fallout from that I guess.

samyron:
> @samyron and others watching this PR. I am sorry that we are in the position of two complete implementations but I think we will be better for it in the end. I am going to continue to examine the impls today and probably tomorrow. I really like the idea of using this one because it is so close to the C version (which is good for synch'ing when new things happen). That said, my parser is doing a little better in most of the benchmarks (twitter and citm_catalog being the exceptions --- and also probably the most real-world examples). I think if I can get this one's speed up to mine then we should use this one.

My apologies to you @enebo! Had I seen your [comment](https://github.com/ruby/json/issues/983#issuecomment-4603423404) I wouldn't have created this one Parser. That said.. I'm totally cool combining efforts in an attempt to get the fastest parser possible.

samyron:
> So something magical is happening in this PRs parser which is not on mine. Yet, most of the benchmarks on mine will edges out this one until it hits multiple byte UTF-8 characters. Then this parser does quite a bit better again (citm_catalog and twitter).

It might be worth an attempt to incorporate the `StringScanner` and `VectorizedStringScanner` into your string parsing function to see if that helps. I can try that later tonight. I'm assuming [this is the branch](https://github.com/ruby/json/compare/master...enebo:json:new_jruby_parser)?

enebo:
@samyron yeah.  That's the branch.    

The phase approach might actually improve my branch as well since I have a lot of state machine switch nonsense making sure I am not in the wrong place in many places.  Seeing that and how I create array/hash at front vs back of element processing really tickles my brain.  Like I spend cost of potentially regrowing an array as I add to it by using a RubyArray right away but the value stack in the PR means walking down it for hash sets where early access to it is probably quicker.  Always fun to see different approaches.

samyron:
> @samyron yeah. That's the branch.
> 
> The phase approach might actually improve my branch as well since I have a lot of state machine switch nonsense making sure I am not in the wrong place in many places. Seeing that and how I create array/hash at front vs back of element processing really tickles my brain. Like I spend cost of potentially regrowing an array as I add to it by using a RubyArray right away but the value stack in the PR means walking down it for hash sets where early access to it is probably quicker. Always fun to see different approaches.

I wonder if this accounts for some of the difference in the float parsing benchmark you mentioned above. Looking at the data in [canada.json](https://github.com/ruby/json/blob/master/benchmark/data/canada.json), there are arrays of arrays of arrays.

Looking at the "second level" array sizes:

```
irb(main):013> coords = data['features'].first['geometry']['coordinates']
...
irb(main):014> coords.each { |c| puts c.length } 
14
33
18
23
10
28
9
28
279
221
11
86
28
19
26
23
24
38
24
```

When this branch is decoding an array, we know exactly how many elements it contains and allocate an array of exactly that size and use `System.arraycopy` to build the `RubyArray`. 

Additionally, the code only resets the top index in the value stack array, so we hit that array with 279 element, we don't need to reallocate the value stack until we have an array with more elements.

enebo:
"When this branch is decoding an array, we know exactly how many elements it contains and allocate an array of exactly that size and use System.arraycopy to build the RubyArray."

Yeah.  That is most definitely an additional cost and it could definitely explain why my faster floating point is quite a bit slower than when I moved it to your parser.  I could definitely special case arrays to use that technique though.  It is just an extra field.  The bigger question to me though is there a benefit for hashes to make a RubyHash upfront?  Like if we waited we could specify a "good" amount of buckets (this PR has no bucket size provided but it could help).  I am less clear on buckets because it becomes a space vs time thing and it is complicated by how much the keys distribute.

enebo:
@samyron I shoe-horned a value array for just making arrays and it did speed up.  How I did it was a little gross (I push a Ruby fixnum instead of a RubyArray up front and use that when I hit ']' to make proper array), but it did not make it to the result on this branch.  I am now at:

```text
 ~/work/json_newest/benchmark new_jruby_parser * 360% ONLY=json ~/work/jruby-10.1/bin/jruby -J-XX:+UseParallelGC -Xcompile.invokedynamic -I../lib parser.rb
== Parsing float parsing (2251051 bytes)
jruby 10.1.1.0-SNAPSHOT (4.0.0) 2026-06-08 482a54b9cb OpenJDK 64-Bit Server VM 25.0.1+8-27 on 25.0.1+8-27 +indy +jit [x86_64-linux]
                json    12.000 i/100ms
                json    142.612 (± 1.4%) i/s    (7.01 ms/i) -      1.428k in  10.016141s

```

Still, it is showing promise.

samyron:
> (this PR has no bucket size provided but it could help).

🤦 Doh! This started out much simpler and I was initially parsing Java objects while I was getting the overall shape of the parser put into place. Objects were a `HashMap` for a while with a capacity of `(int) (size / 0.75f)`. When porting over to Ruby I must have dropped that. I need to dig in to the implementation of `RubyHash` and figure out how to compute the number of buckets needed to avoid resizes.

headius:
Fwiw JRuby 10.1 has been laying the groundwork to start replacing the old linked buckets hash with a smaller more efficient implementation. It may not be super valuable to try to game the buckets but there may be some quick wins.

samyron:
@enebo I cherry picked your parseNumber commit onto a local copy of this PR. It is significantly faster. I can push it here unless you'd like to contribute it directly. 

Additionally, with respect to the other benchmarks... I don't believe your parser does UTF-8 validation. I have a branch that removes that validation. You can find those benchmarks in [this comment above](https://github.com/ruby/json/pull/1004#issuecomment-4714444425).

enebo:
@samyron yeah I say add my number commit to this.   I think we should move forward with your version but I will add PRs later for this.

 I thought about things and I think the performance difference with respect to the benchmarks is that small arrays and hashes favors the upfront creation of hash/array.  Larger ones favors post creation (especially arrays).   I did not try to prove this but it seems fairly clear.  Weirdly, during development of my parser I would reach stages where that small array of numbers would jump from 800k i/s to 1.1m i//s and then a tiny change would drop it back down.  It would only affect just that benchmark but it always made me wonder what thing is not happening. 

So, I could see if JSON were always tiny payloads of small data structures then I get a few percentage points over what you are submitting but I think we have payloads arrays larger than 16 (default array alloc in JRuby RubyArray today) commonly in input then the upfront arrays start performing more arrayCopy operations.  I cannot evaluate hashes but late creation at least gives the ability to pick hashing strategy.

"Additionally, with respect to the other benchmarks... I don't believe your parser does UTF-8 validation. I have a branch that removes that validation. You can find those benchmarks in https://github.com/ruby/json/pull/1004#issuecomment-4714444425."

That could be some part of it but it is not for the two benchmarks from that comment.  Those are the only two benchmarks where mine is slower than this PR.    I think my non-validating string processing is probably slower than your validating one.  The twitter one is almost entirely string processing and it has quite a few mutlibyte character strings in it.

I can see some things we can change like my options processing is going to be faster (not benchmarked yet?) since it only walks number of options provided.  This morning, I will also look at this PR with jitwatch and see if anything stands out.  

I could see making multiple string paths based on the dimensions of key/value and ascii/mbc but that would just be a future experiment.

samyron:
> @samyron yeah I say add my number commit to this. I think we should move forward with your version but I will add PRs later for this.

Done.

> I thought about things and I think the performance difference with respect to the benchmarks is that small arrays and hashes favors the upfront creation of hash/array. Larger ones favors post creation (especially arrays). I did not try to prove this but it seems fairly clear. Weirdly, during development of my parser I would reach stages where that small array of numbers would jump from 800k i/s to 1.1m i//s and then a tiny change would drop it back down. It would only affect just that benchmark but it always made me wonder what thing is not happening.

I found this too, small changes could significantly impact performance in unexpected ways.

Another anomaly I encountered was running a benchmark in isolation like `twitter.json` versus running `activitypub.json` then `twitter.json` could affect the results of the `twitter.json` benchmark. My _guess_ is HotSpot was making some decisions based off of how the code was behaving in the first benchmark but the second was different enough that the assumptions/decisions were no longer that great. However, it wasn't consistent so I'm not entirely sure that was it. Additionally, I had to run my benchmarks for at least 10 seconds for things to get stable. 

> This morning, I will also look at this PR with jitwatch and see if anything stands out.

I would love to know the what this shows. This was on my list of things to do... One thing I was experimenting with a lot last night was if the SWAR-based `StringScanner` is actually helpful. Benchmarking a bit this morning shows a scalar loop might be just as fast... One heuristic I might try later is to switch to a scalar loop for object keys as they are usually short. Not sure if it's worth it though.

enebo:
I did not see anything I didn't expect in jitwatch.  I do have two simple changes I made on enebo/json@one_whitespace: 463e564ba42dec6a5ab28f2a4f90046807c8887a and fb4d1c3fa940fde8ca728639a96ed064a117fdd8.  They do not affect anything but the main run loop just has a single eatWhitespace and I also removed currentNesting.  This does not impact performance but I think it removes some code from run and that is always nice.  Not sure but the C code probably could do the same thing?

Anyways, I am gone until Friday.  I reiterate I am ok with this PR landing over mine.

headius:
Merging now and iterating in place seems like a good strategy to me. I'm holding back doing some profiling and micro-optimizations until things land as well, but we'll want to see how these hot loops and giant switch statements perform on typical JVM JIT before we do anything. I am ok with whatever @samyron and @enebo decide about merging this and moving forward.

samyron:
> Merging now and iterating in place seems like a good strategy to me. I'm holding back doing some profiling and micro-optimizations until things land as well, but we'll want to see how these hot loops and giant switch statements perform on typical JVM JIT before we do anything. I am ok with whatever @samyron and @enebo decide about merging this and moving forward.

I'm good with merging now though I am working on a fast-path in `StringDecoder#decode` when we're decoding ASCII-only strings. This keeps the UTF-8 validation in the case were we detect a non-ASCII byte in a string but doesn't punish ASCII-only strings with backslashes/escapes or control characters. I'm happy to create a PR for this separately after I do a bit more profiling tonight.

byroot:
FYI: I added the nst/JSONTestSuite in https://github.com/ruby/json/pull/1015, which found a couple previously unknown difference between the Java and C parsers. You might want to rebase because I suspect some of these new tests may be failing on your branch, so you'll like have to update the `JRUBY_PENDING` list (hopefully removing items, but it's OK if you need to add some).

samyron:
> FYI: I added the nst/JSONTestSuite in #1015, which found a couple previously unknown difference between the Java and C parsers. You might want to rebase because I suspect some of these new tests may be failing on your branch, so you'll like have to update the `JRUBY_PENDING` list (hopefully removing items, but it's OK if you need to add some).

Done. The new parser successfully parsed all of the `JRUBY_PENDING` tests so I removed it from `json_minefield_parser_test.rb `.

PR #1006, opened 2026-06-15
Replace Ryu with a C99 port of fast_float

byroot:
Contrary to Ryu, it remains correct up to 18 mantissa digits, meaning we don't have to fallback as much to Ruby's much slower `rb_cstr_to_dbl`.

On `canada.json`, the most number heavy benchmark, `ffp_s2d` is less than 5% of total runtime. Most of the time is actually sent parsing, not decoding.

There is a more complete C99 port of `fast_float` at https://github.com/kolemannix/ffc.h which also does parsing and could be worth considering, however it might make it much harder to detect when we're outside the safe range and to fallback, so unclear if it would be a win.

Credit to @tilo for the C99 port (very sligthly reworked by myself).

```
== Parsing citm_catalog.json (500124 bytes)
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +YJIT +PRISM [arm64-darwin25]
               after    73.000 i/100ms
               after    733.301 (± 1.0%) i/s    (1.36 ms/i) -      3.723k in   5.077044s

Comparison:
before:      716.8 i/s
 after:      733.3 i/s - 1.02x  faster

== Parsing float parsing (2090303 bytes)
ruby 4.0.5 (2026-05-20 revision 64336ffd0e) +YJIT +PRISM [arm64-darwin25]
               after    35.000 i/100ms
               after    353.130 (± 0.3%) i/s    (2.83 ms/i) -      1.785k in   5.054791s

Comparison:
before:      323.1 i/s
 after:      353.1 i/s - 1.09x  faster
```


byroot:
Need some fix for windows support: `error C4013: '__builtin_clzll' undefined; assuming extern returning int`

PR #1015, opened 2026-06-18
Vendor tests from https://github.com/nst/JSONTestSuite

byroot:
It's only testing if we run into a parsing error or not, not testing the actual parsed value, but that's still useful to help ensure we're not changing behavior on dark corners of the spec.


test/json/json_minefield_parser_test.rb
```diff
@@ -0,0 +1,114 @@
...
+
+  if RUBY_ENGINE == 'jruby'
+    # The Java parser validate the document encoding
+    # but the C parser doesn't.
+    # See: https://github.com/ruby/json/issues/138
+    UNDEFINED_FAILING.concat(INVALID_ENCODING_TESTS)
+  end
+
+  fixtures.each do |path|
+    payload = File.read(path)
```
nobu, line 81: Lots of encoding errors.
http://rubyci.s3.amazonaws.com/debian/ruby-master/log/20260618T093002Z.fail.html.gz

Probably need to read in binary mode:
```suggestion
    payload = File.binread(path)
```

PR #1016, opened 2026-06-18
Is it intentional that `#partial_value` calls `on_load`?

mame:
Currently `#partial_value` sometimes calls `on_load`. Is this intentional?

If it is, the `in_use` lock has to cover `#partial_value` too: otherwise touching the parser from within `on_load` triggers a use-after-free.  
```ruby
require "json"
parser = nil
parser = JSON::ResumableParser.new(on_load: ->(o) {
  parser.parse if o.is_a?(Array) # re-entrant
  o
})
parser << "[1,"
parser.parse
parser.partial_value # AddressSanitizer: heap-use-after-free
```

The fix in this PR holds `in_use` for the duration of `#partial_value`, like `#parse` does.

Personally, `#partial_value` is interesting, but I am not sure it has a real use case.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -2481,18 +2481,9 @@ static VALUE cResumableParser_clear(VALUE self)
     return self;
 }
 
-/*
- * call-seq: partial_value -> object
- *
- * Returns the Ruby objects parsed up to this point:
- *   parser << '[1, [2, 3,'
- *   parser.parse # => false
- *   parser.value # ArgumentError no ready value
- *   parser.partial_value # => [1, [2, 3]]
- */
-static VALUE cResumableParser_partial_value(VALUE self)
+static VALUE cResumableParser_partial_value_body(VALUE self)
 {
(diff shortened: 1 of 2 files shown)


byroot:
> Currently `#partial_value` sometimes calls `on_load`. Is this intentional?

It is yes.

mame:
I was wondering what would happen if incomplete data were passed to `create_additions` via `on_load`, but it turns out `ResumableParser` doesn't support `create_additions` at all. I think that's fine then.

By the way, passing keyword arguments like `create_additions:` to `ResumableParser.new` silently ignores them, which might be a little confusing.

byroot:
> passing keyword arguments like `create_additions:` to `ResumableParser.new` silently ignores them, which might be a little confusing.

Yes, it's because like all of JSON, there is actually no keyword arguments, it's all option hashes.

I would like to transition to keyword arguments for all of `JSON`, but I think that's something for 3.0.

But perhaps `Resumable` should already use keyword arguments. I'll think about it.

PR #1019, opened 2026-06-18
Fix `JSON::ResumableParser` stalling on a NUL byte

mame:
Feeding `"[\x00]"` makes `#parse` return false forever and never make progress.

```ruby
require "json"
parser << "[\x00]"
parser.parse   # => false
parser.parse   # => false
parser.parse   # => false (forever)
```

Fix: a NUL byte that is not at the end of the stream is treated as a syntax error (it can never become valid by feeding more), as the non-resumable parser already does.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -1642,7 +1642,7 @@ ALWAYS_INLINE(static) bool json_parse_any(JSON_ParserState *state, JSON_ParserCo
                     state->cursor++;
                     value = json_decode_array(state, config, 0);
                     break;
-                } else if (resumable && next == 0) {
+                } else if (resumable && eos(state)) {
                     state->cursor = value_start;
                     return false;
                 }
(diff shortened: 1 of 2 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -1691,6 +1691,12 @@ ALWAYS_INLINE(static) bool json_parse_any(JSON_ParserState *state, JSON_ParserCo
             }
 
             case 0:
+                // peek() returns 0 both at end-of-stream and for a literal NUL byte in the
+                // buffer. Only a genuine EOS means "feed me more"; a NUL byte that is not at
+                // EOS is just an invalid character.
+                if (RB_UNLIKELY(resumable && !eos(state))) {
```
byroot:
```suggestion
                if (RB_UNLIKELY(!eos(state))) {
```

Hum, that's not limited to resumable mode, is it?

mame: Indeed. Thanks!

PR #1020, opened 2026-06-18
Reset all per-parse state in `JSON::ResumableParser#clear`

mame:
Currently, `#clear` does not reset `current_nesting`.

```ruby
require "json"
parser = JSON::ResumableParser.new(max_nesting: 10)
10.times do
  parser << "[1"
  parser.parse
  parser.clear
end
parser << "[1"
parser.parse   # JSON::NestingError: nesting of 11 is too deep
```

This PR resets `current_nesting`.
                                                                                                                                               It also resets `in_array` and `emitted_deprecations` just for case, though I am not sure whether we should reset `emitted_deprecations` or not.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -2487,6 +2487,9 @@ static VALUE cResumableParser_clear(VALUE self)
     parser->frames.head = 0;
     parser->value_stack.head = 0;
     parser->state.name_cache.length = 0;
+    parser->state.current_nesting = 0;
+    parser->state.in_array = 1;
+    parser->state.emitted_deprecations = 0;
     parser->state.start = parser->state.cursor = parser->state.end = NULL;
     return self;
 }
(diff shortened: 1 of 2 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -2478,13 +2477,25 @@ static VALUE cResumableParser_clear(VALUE self)
     parser->frames.head = 0;
     parser->value_stack.head = 0;
     parser->state.name_cache.length = 0;
+    parser->state.current_nesting = 0;
+    parser->state.in_array = 1;
+    parser->state.emitted_deprecations = 0;
```
byroot, line 2492: I wonder if we should just `memset(parser, 0, sizeof(*parser))`.

byroot: I guess not. Because then we need to be careful not to leak the buffers we allocated etc. Nevermind.

PR #1021, opened 2026-06-18
Fix off-by-one line number in parse error messages

mame:
Parse error messages report the line number one too low whenever the error is past the first line.

```ruby
require "json"
JSON.parse("[1,\n@")
# expected: unexpected character: '@' at line 2 column 1
# actual:   unexpected character: '@' at line 1 column 1
```


byroot:
Looks like the same bug is present on JRuby :)

PR #1031, opened 2026-06-19
ResumableParser: use throw rather than raise for handled EOS

byroot:
Since the exception will be swallowed, building a message and backtrace is just a waste of time.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -677,10 +677,17 @@ static VALUE parse_error_new(JSON_ParserState *state, VALUE message, long line, 
 
 NORETURN(static) void raise_parse_error(const char *format, JSON_ParserState *state, bool eos)
 {
-    VALUE message = build_parse_error_message(format, state);
-    if (state->parser) { // line and columns can't be accurate in resumable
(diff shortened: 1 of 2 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -2379,14 +2386,25 @@ static VALUE cResumableParser_feed(VALUE self, VALUE str)
...
+static VALUE json_parse_any_resumable_safe0(RB_BLOCK_CALL_FUNC_ARGLIST(yielded_arg, _args))
 {
     struct json_parse_any_args *args = (struct json_parse_any_args *)_args;
     return (VALUE)json_parse_any(args->state, args->config, true);
 }
 
+static VALUE json_parse_any_resumable_safe(VALUE _args)
+{
+    struct json_parse_any_args *args = (struct json_parse_any_args *)_args;
+    VALUE result = rb_catch_obj(args->parser, json_parse_any_resumable_safe0, _args);
```
byroot: It looks like on TruffleRuby, `rb_catch_obj` doesn't call `json_parse_any_resumable_safe0` at all. Unclear why...

PR #1032, opened 2026-06-19
ResumableParser: accept only keyword arguments

byroot:
Fix: https://github.com/ruby/json/pull/1016#issuecomment-4744487710

`json` takes option hashes across the board, mostly because its API predates the introduction of keyword arguments.

I'd like to change that to only take keyword arguments and error when an unknown argument is passed, but I'm not yet sure of the backward compatibility consequences, so it might wait for the next major.

But in the meantime, `ResumableParser` being a new API, it can safely use keyword arguments.

FYI: @mame

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -1966,6 +1966,8 @@ static VALUE convert_encoding(VALUE source)
 struct parser_config_init_args {
     JSON_ParserConfig *config;
     VALUE self;
+    VALUE unknown_keywords;
+    bool strict;
 };
 
 static void parser_config_wb_write(VALUE self, VALUE *dest, VALUE val)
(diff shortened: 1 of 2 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -2019,27 +2021,43 @@ static int parser_config_init_i(VALUE key, VALUE val, VALUE data)
...
+    rb_hash_foreach(opts, parser_config_init_i, (VALUE)&args);
 
+    if (RB_UNLIKELY(args.unknown_keywords)) {
+        if (RARRAY_LEN(args.unknown_keywords) == 1) {
+            rb_raise(rb_eArgError, "unknown keyword: %" PRIsVALUE, RARRAY_AREF(args.unknown_keywords, 0));
+        }
+        else {
+            VALUE keywords = rb_ary_join(args.unknown_keywords, rb_utf8_str_new_cstr(", "));
+            rb_raise(rb_eArgError, "unknown keywords: %s", RSTRING_PTR(keywords));
+            RB_GC_GUARD(keywords);
```
nobu, line 2059: Why not using `PRIsVALUE` here as well as the `if` side?

byroot: No particular reason, I just never think of `PRIsVALUE`, you're right it would be simpler.

PR #1035, opened 2026-06-21
Cleanup the `rb_catch_obj` workaround

byroot:
Followup: https://github.com/ruby/json/pull/1031

Now that the bug has been fixed upstream, we can only apply it on unpatched version.

Also refactor the workaround to expose the exact same API as `rb_catch_obj`, making it easier to ignore and remove later.

### DIFF
--- a/ext/json/ext/parser/extconf.rb
+++ b/ext/json/ext/parser/extconf.rb
@@ -2,7 +2,12 @@
 require 'mkmf'
 
 $defs << "-DJSON_DEBUG" if ENV.fetch("JSON_DEBUG", "0") != "0"
-$defs << "-DJSON_WORKAROUND_RB_CATCH_BUG" if RUBY_ENGINE == 'truffleruby'
+
+if RUBY_ENGINE == 'truffleruby' && RUBY_VERSION < '4.0'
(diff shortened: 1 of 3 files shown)


ext/json/ext/parser/extconf.rb
```diff
@@ -2,7 +2,12 @@
 require 'mkmf'
 
 $defs << "-DJSON_DEBUG" if ENV.fetch("JSON_DEBUG", "0") != "0"
-$defs << "-DJSON_WORKAROUND_RB_CATCH_BUG" if RUBY_ENGINE == 'truffleruby'
+
+if RUBY_ENGINE == 'truffleruby' && RUBY_VERSION < '4.0'
```
byroot, line 6: FYI @eregon, let me know if the logic ain't correct here, I'm not very familiar with Truffle

eregon: That works (in both cases the workaround is not applied on truffleruby-head which is now fixed) but I think `RUBY_ENGINE_VERSION` is little bit cleaner (explicitly mentions the version fixing the bug)
```suggestion
if RUBY_ENGINE == 'truffleruby' && RUBY_ENGINE_VERSION < '40.0'
```
https://github.com/ruby/json/pull/1036

Issue #976, opened 2026-04-20
`::JSON.pretty_generate` should sort hash keys


jcpunk:
**Summary**
`JSON.pretty_generate` currently preserves the insertion order of Ruby hashes. This leads to non-deterministic output when hash construction order varies, making diffs noisy and reproducibility harder. Add an option to sort hash keys during generation to produce stable, predictable JSON output.

**Motivation / Problem**

* In many workflows (config generation, test fixtures, CI artifacts), stable serialization is critical.
* Hash insertion order may differ across code paths, Ruby versions, or data sources, causing semantically identical objects to produce different JSON.
* This complicates:

  * Git diffs and code reviews
  * Caching and content hashing
  * Snapshot testing
  * Reproducible builds

**Proposed Solution**
Introduce an option to `JSON.pretty_generate` (and possibly `JSON.generate`) to sort object keys lexicographically.

**API Options (one of):**

1. Keyword argument:

   ```ruby
   JSON.pretty_generate(obj, sort_keys: true)
   ```
2. Extend `JSON::State`:

   ```ruby
   state = JSON::State.new(sort_keys: true)
   JSON.pretty_generate(obj, state)
   ```

**Behavior**

* When `sort_keys: true`, all hashes are serialized with keys sorted (string comparison).
* Default remains `false` to preserve backward compatibility and performance characteristics.

**Example**

```ruby
obj = { b: 1, a: 2 }

JSON.pretty_generate(obj)
# => {
#      "b": 1,
#      "a": 2
#    }

JSON.pretty_generate(obj, sort_keys: true)
# => {
#      "a": 2,
#      "b": 1
#    }
```

**Alternatives Considered**

* Pre-sorting hashes before serialization:

  * Requires deep traversal and duplication of data structures
  * Error-prone and inefficient for large nested objects

* Relying on insertion order discipline:

  * Not robust across boundaries or contributors

**Impact**

* Improves determinism and reproducibility across tooling and environments
* Reduces diff noise and improves developer experience
* Aligns with behavior available in other ecosystems (e.g., Python’s `json.dumps(sort_keys=True)`)

**Performance Considerations**

* Sorting introduces overhead proportional to key count per object
* Acceptable when opt-in; no impact on default behavior

**Backward Compatibility**

* Fully backward compatible if default remains unsorted

**Test Plan**

* Unit tests verifying:

  * Sorted vs unsorted output for flat and deeply nested hashes
  * Stability across multiple invocations
[...]

Issue #1038, opened 2026-06-24
Add support for JSON Canonicalization (RFC8785)


renchap:
There is an existing Gem implementation for [RFC8785](https://datatracker.ietf.org/doc/html/rfc8785) (JSON Canonicalization): [`json-canonicalization`](https://github.com/dryruby/json-canonicalization)) but unfortunately its maintainer passed away.

We wonder if this could be supported directly by the `json` gem.

This RFC is important when you want to use cryptographic signatures with JSON documents, as the signature depends on the keys' order and format. In our case, we need this to sign ActivityPub JSON objects in Mastodon.

The above implementation's code is quite simple and it has good test coverage, so it could be used as a starting point?

This could also solve #976 in a standard way.

datibbaw:
What would you expect the API to look like? Perhaps something like this?
```ruby
require "json"

JCS_KEY_SORT = lambda do |hash|
  hash.sort_by do |key, _value|
    raise TypeError, "JCS object keys must be strings" unless key.is_a?(String)

    key.encode(Encoding::UTF_16BE).unpack("n*")
  end.to_h
end

JCS = JSON::Coder.new(
  sort_keys: JCS_KEY_SORT,
  number_serializer: ECMAScriptNumbers.method(:serialize) # imaginary method that converts numbers according to the spec
)

canonical_json = JCS.dump(value)
```

byroot:
What I had in mind was more: `JSON.dump(obj, rfc8785: true)` or of course `JSON::Coder.new(rfc8785: true)`. At least for the user side of things.

Internally, that could just provide a Ruby block to invoke with numbers to do the serialization, that's if we don't want to bother implementing the logic 3 times, but ideally the generators would know how to do this. (The fact that there is 3 generators implementation is largely why I have procrastinated on this).

Alternatively, we could also potentially always serialize numbers "canonically" that's something I need to look at, but IIRC since they chose to have capital `E` etc, that may seem weird to users.

datibbaw:
> Alternatively, we could also potentially always serialize numbers "canonically" that's something I need to look at, but IIRC since they chose to have capital E etc, that may seem weird to users

Not just that 😬 it would also change the representation e.g.
```
1.0    -> 1
-0.0   -> 0
1e15   -> 1000000000000000
1.0E30 -> 1e+30
```

Pushing the `sort_keys` proc down with `rfc8785: true` looks reasonable ... updating three generators does seem like a fair amount of work; `fpconv` helps to a degree for cruby at least 😸

PR #1043, opened 2026-06-25
Add `sort_keys` option in the generator.

samyron:
I had some time last night so I decided to see how hard it would be to implement a [sort_keys](https://github.com/ruby/json/issues/976) option to the generator. This adds support to the C extension, Java extension and truffleruby implementation.

```
irb(main):005> h = { zzz: true, a: 1, c: 2, b: "Test", 'z': 'blah 2'} 
=> {zzz: true, a: 1, c: 2, b: "Test", z: "blah 2"}
irb(main):006> JSON.generate(h, sort_keys: false)
=> "{\"zzz\":true,\"a\":1,\"c\":2,\"b\":\"Test\",\"z\":\"blah 2\"}"
irb(main):007> JSON.generate(h, sort_keys: true)
=> "{\"a\":1,\"b\":\"Test\",\"c\":2,\"z\":\"blah 2\",\"zzz\":true}"
irb(main):008> JSON.generate(h)
=> "{\"zzz\":true,\"a\":1,\"c\":2,\"b\":\"Test\",\"z\":\"blah 2\"}"
```

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -34,13 +34,14 @@ typedef struct JSON_Generator_StateStruct {
     bool ascii_only;
     bool script_safe;
     bool strict;
+    VALUE sort_keys;
 } JSON_Generator_State;
 
-static VALUE mJSON, cState, cFragment, eGeneratorError, eNestingError, Encoding_UTF_8;
+static VALUE mJSON, cState, cFragment, eGeneratorError, eNestingError, Encoding_UTF_8, default_sort_keys_proc;
 
 static ID i_to_s, i_to_json, i_new, i_encode;
 static VALUE sym_indent, sym_space, sym_space_before, sym_object_nl, sym_array_nl, sym_max_nesting, sym_allow_nan, sym_allow_duplicate_key,
-             sym_ascii_only, sym_depth, sym_buffer_initial_length, sym_script_safe, sym_escape_slash, sym_strict, sym_as_json;
+             sym_ascii_only, sym_depth, sym_buffer_initial_length, sym_script_safe, sym_escape_slash, sym_strict, sym_as_json, sym_sort_keys;
 
 
 #define GET_STATE_TO(self, state) \
@@ -709,6 +710,7 @@ static void State_mark(void *ptr)
     rb_gc_mark_movable(state->object_nl);
     rb_gc_mark_movable(state->array_nl);
     rb_gc_mark_movable(state->as_json);
+    rb_gc_mark_movable(state->sort_keys);
 }
 
 static void State_compact(void *ptr)
@@ -720,6 +722,7 @@ static void State_compact(void *ptr)
     state->object_nl = rb_gc_location(state->object_nl);
     state->array_nl = rb_gc_location(state->array_nl);
     state->as_json = rb_gc_location(state->as_json);
+    state->sort_keys = rb_gc_location(state->sort_keys);
 }
 
 static size_t State_memsize(const void *ptr)
@@ -769,6 +772,7 @@ static void vstate_spill(struct generate_json_data *data)
     RB_OBJ_WRITTEN(vstate, Qundef, state->object_nl);
     RB_OBJ_WRITTEN(vstate, Qundef, state->array_nl);
     RB_OBJ_WRITTEN(vstate, Qundef, state->as_json);
+    RB_OBJ_WRITTEN(vstate, Qundef, state->sort_keys);
 }
 
 static inline VALUE json_call_to_json(struct generate_json_data *data, VALUE obj)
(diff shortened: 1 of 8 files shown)


byroot:
THanks for handling this, however in lieu of https://github.com/ruby/json/issues/1038, I think a `sort_key` option shouldn't be a a simple boolean, but should also accept a proc so we can implement custom sorting semantic.

samyron:
> THanks for handling this, however in lieu of #1038, I think a `sort_key` option shouldn't be a a simple boolean, but should also accept a proc so we can implement custom sorting semantic.

No problem. I can try to take a stab at that tonight night/this weekend. I _think_ it should be as simple as accepting a proc and passing it to `obj.sort_by` when generating the JSON.

samyron:
I added the ability to pass a Proc which accepts two `[key, value]` pairs to allow for arbitrary sorting. I'm not entirely sure if this is what you had in mind but it allows for maximum flexibility. I'm not totally sure if that's warranted but I ran with that assumption. I'm happy to make changes as requested. 

Note that a boolean (`true`) is still supported which will sort keys lexicographically. 

```
irb(main):002> reverse = ->(a, b) { b[0] <=> a[0] }
=> #<Proc:0x000000012d2bbc08 (irb):2 (lambda)>
irb(main):003> JSON.generate({2=>"a", 1=>"b", 3=>"c"}, sort_keys: reverse)
=> "{\"3\":\"c\",\"2\":\"a\",\"1\":\"b\"}"
```

ext/json/ext/generator/generator.c
```diff
@@ -1050,6 +1054,17 @@ static inline long increase_depth(struct generate_json_data *data)
...
 {
+    if (RB_UNLIKELY(RTEST(data->state->sort_keys))) {
+        VALUE sort_keys = data->state->sort_keys;
+        VALUE sorted_array;
+        if (rb_obj_is_proc(sort_keys)) {
+            sorted_array = rb_funcall_with_block(obj, rb_intern("sort"), 0, NULL, sort_keys);
+        } else {
+            sorted_array = rb_funcall(obj, rb_intern("sort"), 0);
+        }
+        obj = rb_funcall(sorted_array, rb_intern("to_h"), 0);
```
byroot: Sorry, for not immediately communicating in details what I had in mind, I was traveling.

Based on the type of sorting `json-normalization` needs: https://github.com/dryruby/json-canonicalization/blob/a85e688a4a593ed6b28e5aa4610c996828c13ada/lib/json/canonicalization.rb#L64-L78

I think it would be better if the sort proc would just take the entire hash at once, and return a sorted hash.

As for the `sort: true` argument, that's great, that's what I had in mind, but we might be better to just convert it into a proc early on the Ruby side:

```ruby
def generate(..., options)
  if options[:sort_keys] == true
    options = options.merge(sort_keys: ->(hash) { hash.sort.to_h }
  end
  ...
end
```

samyron: I like the idea of the `boolean -> proc` resolution but I don't think there is a single location in Ruby that would cover all of the places where a `boolean` could be passed in as an option.

There are tests where options are passed directly to `JSON::State.new(...)`. If `sort_keys` were passed in we'd need to perform the same `boolean -> proc` resolution. 

If I'm tracing the codepaths correctly:

`JSON.generate -> State.generate (cState_m_generate in C) -> cState_m_do_generate -> configure_state -> configure_state_i`.

Then there is also:

`JSON::State.new(...) -> State#initialize -> State#configure -> _configure -> configure_state -> configure_state_i` then `generate` can be called on this `JSON::State` object directly, which is defined in C.

Finally, `state.sort_keys=` (`cState_sort_keys_set`) could be set directly too. 

One thing I think we can do is define a `SORT_KEYS_PROC` in Ruby, in the `JSON` module:

```
SORT_KEYS_PROC = ->(hash) { hash.sort.to_h }
```

We could create a helper in C to resolve `boolean (true) -> proc` in both `configure_state_i` and `cState_sort_keys_set`.

samyron: I'm happy to change the proc so it receives the full `Hash`.

ext/json/ext/generator/generator.c
```diff
@@ -1050,6 +1054,11 @@ static inline long increase_depth(struct generate_json_data *data)
 
 static void generate_json_object(FBuffer *buffer, struct generate_json_data *data, VALUE obj)
 {
+    if (RB_UNLIKELY(RTEST(data->state->sort_keys))) {
+        obj = rb_funcall(data->state->sort_keys, i_call, 1, obj);
```
byroot:
```suggestion
        obj = rb_proc_call_with_block(data->state->sort_keys, 1, &obj, Qnil);
```

Rather than going through `funcall` which requires a method lookup, we can use `b_proc_call*`

lib/json/common.rb
```diff
@@ -243,6 +243,10 @@ def self.create_id
 
   MinusInfinity = -Infinity
 
+# The default proc used when the +sort_keys+ generation option is +true+.
+# It returns a new hash with the entries sorted by their keys.
+  SORT_KEYS_PROC = ->(hash) { hash.sort.to_h }
```
byroot:
```suggestion
  SORT_KEYS_PROC = ->(hash) { hash.sort.to_h }
  private_constant :SORT_KEYS_PROC
```

byroot: Well, not possible because we have to access it from C, but at least `:nodoc:`

PR #1048, opened 2026-07-03
Add JSON::ResumableParser#partial_value? and #empty?

### LINKED ISSUE #1047: [Question] ResumableParser: recommended way to detect a truncated stream at EOF?
In Fluentd we are migrating our JSON stream parsing from yajl-ruby to `JSON::ResumableParser` (fluent/fluentd#5383). One behavior we want to preserve: when an input stream ends in the middle of a JSON document, the consumer should be able to notice it (Yajl raised a `ParseError` in that situation; we now emit a warning instead of silently dropping the incomplete tail).

## Problem

There doesn't seem to be a single API that answers "did the stream end mid-document?". Our first attempt used `rest.empty?`, but `rest` only reflects unconsumed tokenizer bytes, so a stream cut on a complete token boundary (right after a `:` or `,`) leaves `rest` empty and `eos?` returning `true`, even though a document is clearly in progress:

```ruby
parser << '{"a":1,'
parser.value while parser.parse
parser.rest          # => ""
parser.eos?          # => true    (but the document is incomplete and gets silently dropped)
parser.partial_value # => {"a"=>1}
```

## Reproduction

```ruby
require 'bundler/inline'
gemfile do
  source 'https://rubygems.org'
  gem 'json', '= 2.20.0'
end

CASES = [
  ['{"a":1}',           false],  # [input, truncated?]
  ['{"a":1}{"b":2}',    false],
  ['{"a":1} ',          false],
  ['{"a":1}{"b":2',     true],   # cut inside a number token
  ['{"a":1}{"b":',      true],   # cut right after a colon (token boundary)
  ['{"a":1}{',          true],   # cut right after object open
  ['{"a":1,',           true],   # cut right after a comma (token boundary)
]

CASES.each do |input, truncated|
  parser << input
  parser.value while parser.parse
  detected = !parser.rest.empty? || !parser.partial_value.nil?
  puts "%-18s rest=%-6s eos?=%-5s partial_value=%-12s detected=%s%s" % [
    input.inspect, parser.rest.inspect, parser.eos?,
    parser.partial_value.inspect, detected,
    detected == truncated ? "" : "  <-- MISMATCH"
  ]
end
```

Output on json 2.20.0 (on Ruby 4.0.5):

```
"{\"a\":1}"          rest=""     eos?=true  partial_value=nil          detected=false
"{\"a\":1}{\"b\":2}" rest=""     eos?=true  partial_value=nil          detected=false
"{\"a\":1} "         rest=""     eos?=true  partial_value=nil          detected=false
"{\"a\":1}{\"b\":2"  rest="2"    eos?=false partial_value={"b"=>nil}   detected=true
"{\"a\":1}{\"b\":"   rest=""     eos?=true  partial_value={"b"=>nil}   detected=true
"{\"a\":1}{"         rest="{"    eos?=false partial_value=nil          detected=true
"{\"a\":1,"          rest=""     eos?=true  partial_value={"a"=>1}     detected=true
```

## What we ended up with

`rest` and `partial_value` turn out to be complementary (`rest` catches the mid-token and just-after-`{` cuts, `partial_value` catches the token-boundary cuts), so we currently use:

```ruby
truncated = !parser.rest.empty? || !parser.partial_value.nil?
```

## Questions

1. Is this combination the intended way to detect truncation, and is it something we can rely on across versions? Both behaviors were established empirically, and building a `partial_value` object just for a nil check feels like a workaround.
2. Is `eos?` returning `true` while a document is structurally in progress (`'{"a":1,'`) the intended semantics? If `eos?` is meant to be tokenizer-level only, a doc note would help; if it's meant to answer "is the parser at a clean end of stream?", this looks like it should return `false`.
3. Would you be open to a dedicated predicate (e.g. `incomplete?` / `in_document?`) for this?

Watson1978:
Fixes #1047

#### `partial_value?`

Returns whether a document is currently under construction (unclosed container, key awaiting its value, etc.). It answers the same question as `!partial_value.nil?`, but as a cheap predicate on the parser's internal value stack, without materializing the partially parsed Ruby objects.
A fully parsed document whose value hasn't been retrieved yet is *not* under construction, that state is covered by `value?`.

Note that a container whose first key or element hasn't been parsed yet (e.g. a stream ending right after `{`) has nothing registered on the parser's stacks so `partial_value?` is `false` there, consistently with `partial_value` returning `nil`. 
The truncation is still observable through the buffer: `eos?` is `false` and `rest` isn't empty. The test suite asserts `partial_value? == !partial_value.nil?` across all cases.

#### `empty?`

Strict semantics: `true` only when the buffer is fully consumed, no document is under construction, and no parsed value awaits retrieval.
```ruby
def empty?
  eos? && !partial_value? && !value?
end
```

This is the single call answering the truncation question from #1047:

| Input (after draining with `parse`/`value`) | `empty?` | |
|---|---|---|
| `''` | `true` | nothing fed |
| `'{"a":1}'` | `true` | |
| `'{"a":1}{"b":2}'` | `true` | |
| `'{"a":1} '` | `true` | trailing whitespace |
| `'{"a":1}{"b":2'` | `false` | inside a number token |
| `'{"a":1}{"b":'` | `false` | right after a colon (token boundary) |
| `'{"a":1}{'` | `false` | right after an object open |
| `'{"a":1,'` | `false` | right after a comma (token boundary) |
| `'"abc'` | `false` | inside a string token |
| `'[1,2'` | `false` | unclosed array |

As raised in review, `empty?` is meaningful mid-drain too: after feeding `'{"a":1}{"b":2}'` and a single successful `parse`, `empty?` is `false` (a document remains in the buffer, and the parsed value hasn't been retrieved);
it only turns `true` once both documents are parsed and retrieved. Likewise a fully parsed but unretrieved value keeps `empty?` `false`.

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -2738,6 +2738,41 @@ static VALUE cResumableParser_eos_p(VALUE self)
 }
 
 /*
+ * call-seq: partial_value? -> true or false
+ *
+ * Returns whether a document is currently under construction: an unclosed
+ * container, a key awaiting its value, etc.
+ *
+ * It answers the same question as <tt>!partial_value.nil?</tt>, but as a
+ * cheap predicate on the parser's internal state, without materializing the
+ * partially parsed Ruby objects:
+ *   parser << '{"a":1,'
+ *   parser.parse # => false
+ *   parser.partial_value? # => true
+ *
+ * A fully parsed document whose value hasn't been retrieved yet is not under
+ * construction: #value? returns true and #partial_value? returns false.
+ */
+static VALUE cResumableParser_partial_value_p(VALUE self)
+{
+    JSON_ResumableParser *parser = cResumableParser_get(self);
+
+    // Mirror of #value?: values on the stack while the document isn't DONE
+    // belong to a partially built document. A container whose first key or
+    // element hasn't been parsed yet has no frame nor value registered (the
+    // tokenizer rewinds to the container start on EOS), so that state is
+    // observable through the buffer (#eos?/#rest) instead, keeping this
+    // predicate consistent with #partial_value returning nil.
+    if (parser->value_stack.head > 0) {
+        json_frame *frame = json_frame_stack_peek(&parser->frames);
+        if (frame->phase != JSON_PHASE_DONE) {
+            return Qtrue;
+        }
+    }
+    return Qfalse;
+}
+
+/*
  * call-seq: parsed_bytes -> integer
  *
  * Returns the number of bytes parsed since the start of the current partial value.
@@ -2793,6 +2828,7 @@ void Init_parser(void)
     rb_define_method(cResumableParser, "value", cResumableParser_value, 0);
     rb_define_method(cResumableParser, "value?", cResumableParser_value_p, 0);
     rb_define_method(cResumableParser, "partial_value", cResumableParser_partial_value, 0);
+    rb_define_method(cResumableParser, "partial_value?", cResumableParser_partial_value_p, 0);
     rb_define_method(cResumableParser, "clear", cResumableParser_clear, 0);
     rb_define_method(cResumableParser, "rest", cResumableParser_rest, 0);
     rb_define_method(cResumableParser, "eos?", cResumableParser_eos_p, 0);
--- a/lib/json/ext.rb
+++ b/lib/json/ext.rb
@@ -41,5 +41,31 @@ module JSON
     end
   end
 
+  if defined?(ResumableParser) # Not yet available on JRuby
+    class ResumableParser
+      # Returns whether the parser is entirely done: no unconsumed bytes in
+      # the buffer, no document under construction and no parsed value
+      # awaiting retrieval.
+      #
+      # The main use case is detecting a truncated stream once the input is
+      # exhausted:
+      #
+      #   loop do
+      #     begin
+      #       parser << socket.readpartial(4096)
+      #     rescue EOFError
+      #       break
+      #     end
+      #     while parser.parse
+      #       process(parser.value)
+      #     end
+      #   end
+      #   warn "stream was truncated" unless parser.empty?
+      def empty?
+        eos? && !partial_value? && !value?
+      end
+    end
+  end
+
   JSON_LOADED = true unless defined?(JSON::JSON_LOADED)
 end
--- a/test/json/resumable_parser_test.rb
+++ b/test/json/resumable_parser_test.rb
@@ -239,6 +239,89 @@ class JSONResumageParserTest < Test::Unit::TestCase
     assert_predicate @parser, :eos?
   end
 
+  def test_empty_predicate
+    # empty? is defined on the state left after parsing everything that
+    # could be parsed from the fed bytes, so drain with parse/value first.
+    {
+      ''               => true,  # nothing fed: vacuously empty
+      '{"a":1}'        => true,
+      '{"a":1}{"b":2}' => true,
+      '{"a":1} '       => true,  # trailing whitespace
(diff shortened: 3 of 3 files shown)


ext/json/ext/parser/parser.c
```diff
@@ -2737,6 +2737,51 @@ static VALUE cResumableParser_eos_p(VALUE self)
     return eos(&parser->state) ? Qtrue : Qfalse;
 }
 
+/*
+ * call-seq: incomplete? -> true or false
+ *
+ * Returns whether the stream ends in the middle of a document.
```
byroot: I think your implementation does a bit more than that. Because it also returns true if there is more left to parse in the buffer:

```ruby
parser << '{"a":1}{"b":2}'
parser.parse # => true
parser.incomplete? # => true
parser.parse # => true
parser.incomplete? # => false
```

Watson1978: Good catch!
I think this is inherent to a predicate-style API rather than an implementation detail. At the point of your example, the buffer holds `'{"b":2}'` as raw undecoded bytes; whether that remainder is a complete document or a truncated one cannot be known without parsing it, which is exactly what `parse` does. So a side-effect-free predicate can only give a meaningful truncation answer once `parse` has returned `false` (i.e., after the usual `while parser.parse` drain loop). At that point, no complete document can remain buffered, and the predicate coincides exactly with "the stream was cut mid-document".

I see two ways forward.

1. Keep the current behavior, define the semantics explicitly as "there are unconsumed bytes or a partially built document", document that truncation detection requires draining first, and pin your example in a test as expected behavior. I'm also open to renaming if `incomplete?` reads misleading mid-drain.
2. Replace the predicate with an explicit end-of-stream signal, e.g. `parser.finish`, which raises `ParserError` if a document is in progress (same shape as yajl's `complete_parse` or `Zlib::Inflate#finish`). This removes the timing ambiguity structurally, at the cost of forcing exception handling on callers that only want to log a warning.

I have a mild preference for (1) since our use case (and I suspect most) calls this after the drain loop anyway, but happy to go either way.
Which shape would you prefer?

byroot: I think it's mostly a naming concern. I'm trying to think of a better alternative.

We can probably also define `partial_value?` as an efficient way to know if the parser stopped midway. (because using `#partial_value` requires to build the object graph, which is costly).

That would be more flexible as you could check something like:

```ruby
def incomplete?
  parser.partial_value? || !parser.eof?
end
```

But I think ultimately the semantic most people would want is something like:

```ruby
def complete?
  parser.eof? && !parser.partial_value?
end
```

Which is a less ambiguous way to express:
  - There is nothing less to parse
  - There is no bytes unaccounted for.

byroot:
What about `empty?`, I think it's a bit less ambiguous about its meaning. no?

Watson1978:
`empty?` works for me — read as a pure state predicate ("nothing pending inside the parser") it's arguably cleaner than `complete?`, which smuggles in a notion of "the stream ended".

One edge it forces us to decide explicitly: a parsed-but-unretrieved value (`parse` returned true, `value` not yet called). For `complete?` semantics I'd have said true there; but a parser holding an unretrieved value doesn't read as "empty", so if we go with `empty?` I'd define it strictly: no unconsumed bytes, no partial document, **and** no unretrieved value. For the EOF-after-drain call site this makes no difference, and the strict version keeps the name honest. Does that match your intent?

byroot:
> Does that match your intent?

I think so yes.

Also ideally your PR only defines `partial_value?` in C, and `empty?` can be define in Ruby, making it easier for users to read.

Watson1978:
Thanks

byroot:
Welcome, thanks for the feedback on the API and the patch.

PR #1053, opened 2026-07-13
Rely on Ruby's HAVE_BUILTIN___BUILTIN_CLZLL

byroot:
We don't need to perform out own checks.

Fix: https://github.com/ruby/json/pull/1052


nobu:
Gemini suggests that “out own checks” might be a typo for “our own checks.”

byroot:
It's right, but since it's in a commit message, there isn't much I can do anymore.

nobu:
Yes, nothing we can for this repo, but added the `log-fix` notes to ruby/ruby.

```console
git notes --ref=log-fix add -m '3s/out/our/' 5a6cbaa2913
```

PR #1062, opened 2026-08-17
Add JSON::ParserError#json_path

### LINKED ISSUE #954: Subtle breaking change with custom `object_class` and duplicate keys
I have a codebase in which we reject duplicate JSON keys with a friendly error message. We achieved this by parsing with something like:

```ruby
JSON.parse(json_string, object_class: TrackingHash, array_class: TrackingArray)
```

Then, everywhere `[]=` is called on `TrackingHash` we track if the key has been added before. If it is duplicated, we add it to a list of duplicated keys. We then can reconstruct (in a single parse) errors like:

```ruby
[
  "a.b.c is a duplicate key",
  "a.d.1.dup is a duplicate key",
  "a is a duplicate key",
]
```

As of sometime recently (~2.13 ? I haven't bisected the exact version but it doesn't work on latest and it did work on 2.9.1) the second `[]=` is never invoked on the custom class. I did discover that you can set `allow_duplicate_key: false` and have it raise, but I can't reconstruct the keypath--so there's no way to tell if `"c"` is duplicated at the root or somewhere nested and I would ideally like to know where in the JSON this occurs.

I imagine this is fairly niche but it would be nice if either the old behavior was restored so I could track this myself or if the error message had the whole keypath in it so I could at least parse it out. In that world I'd still only be able to detect one duplicate at a time but that would be more or less fine. I'd like to upgrade to the latest version (lots of nice improvements in perf, etc.!) but I have to either change this behavior in my app or stick on the version I'm on.

Maybe this isn't a big enough issue to fix but it is a somewhat subtle and confusing change. I would expect `allow_duplicate_key: true` at least to still provide it to my custom class otherwise there's no way to access it (that I am aware of).

ojuschugh1:
Implements the `ParserError#json_path` idea from #954: a JSONPath-style string locating a parse error in the document.

```ruby
begin
  JSON.parse('{"user": {"roles": [1, {"admin": xyz}]}}')
rescue JSON::ParserError => e
  e.message   # => "unexpected character: 'xyz}]}}' at line 1 column 34"
  e.json_path # => "$.user.roles[1].admin"
end

JSON.parse('{"x": {"a": 1, "a": 2}}', allow_duplicate_key: false)
# => ParserError, json_path == "$.x.a"
```

The path is reconstructed lazily at raise time: the frame stack still holds every enclosing container at that point and their keys are still on the rvalue stack, so nothing is added to the happy path. `benchmark/parser.rb` before/after shows no difference outside noise.

A few choices worth reviewing:

- For duplicate key errors the path points at the duplicated key itself (`$.x.a`) rather than the containing object, which seemed closer to what the reporter needs. Line/column keep pointing at the object's opening brace, so the two diverge slightly there. Happy to make it point at the object instead if you'd rather keep them consistent.
- Keys that aren't plain identifiers use bracket notation with escaping: `$["hello world"]`, `$["a\"b"]`. `symbolize_names` and `on_load`-transformed keys are handled.
- It also works for `ResumableParser`, where line/column can't be accurate: an error raised mid-feed still reports the exact path.
- `cParser_parse` released the spilled stacks before raising the end-of-input error. That was fine while nothing read parser state during the raise, but the path reconstruction does (the 100k-deep minefield fixture segfaulted), so the raise now happens before the release. The existing comment already documents that skipping the release on the exception path doesn't leak.
- Not implemented for the Java parser, `json_path` returns nil there like `line`/`column` do. Can look at that as a follow-up if there's interest.

Closes #954

### DIFF
--- a/ext/json/ext/parser/parser.c
+++ b/ext/json/ext/parser/parser.c
@@ -651,60 +651,30 @@ static VALUE build_parse_error_message(const char *format, JSON_ParserState *sta
     return rb_enc_sprintf(enc_utf8, format, ptr);
 }
 
-static bool json_path_append_key(VALUE path, VALUE key)
-{
-    if (RB_SYMBOL_P(key)) {
-        key = rb_sym2str(key);
-    } else if (!RB_TYPE_P(key, T_STRING)) {
-        return false;
-    }
-
-    long len = RSTRING_LEN(key);
-    bool plain = len > 0;
-    for (long i = 0; plain && i < len; i++) {
-        char c = RSTRING_PTR(key)[i];
-        bool alpha = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_';
-        plain = alpha || (i > 0 && c >= '0' && c <= '9');
-    }
-
-    if (plain) {
-        rb_str_cat_cstr(path, ".");
-        rb_str_cat(path, RSTRING_PTR(key), len);
-        return true;
-    }
-
-    rb_str_cat_cstr(path, "[\"");
-    for (long i = 0; i < RSTRING_LEN(key); i++) {
-        char c = RSTRING_PTR(key)[i];
-        if (c == '"' || c == '\\') {
-            rb_str_cat(path, "\\", 1);
-        }
-        rb_str_cat(path, &c, 1);
-    }
-    rb_str_cat_cstr(path, "\"]");
-    return true;
-}
-
 static VALUE json_path_new(JSON_ParserState *state, VALUE duplicate_key)
 {
-    VALUE path = rb_utf8_str_new("$", 1);
+    VALUE path = rb_ary_new_capa(state->current_nesting);
+
     json_frame_stack *frames = state->frames;
     rvalue_stack *values = state->value_stack;
 
     for (long depth = 1; depth < frames->head; depth++) {
         json_frame *frame = &frames->ptr[depth];
+
         bool innermost = depth == frames->head - 1;
         long child_head = innermost ? values->head : frames->ptr[depth + 1].value_stack_head;
         long count = child_head - frame->value_stack_head;
 
         if (frame->type == JSON_FRAME_ARRAY) {
-            rb_str_catf(path, "[%ld]", frame->phase == JSON_PHASE_ARRAY_COMMA ? count - 1 : count);
+            rb_ary_push(path, LONG2NUM(frame->phase == JSON_PHASE_ARRAY_COMMA ? count - 1 : count));
         } else if (innermost && !UNDEF_P(duplicate_key)) {
-            if (!json_path_append_key(path, duplicate_key)) break;
+            rb_ary_push(path, duplicate_key);
         } else if (count & 1) {
-            if (!json_path_append_key(path, values->ptr[child_head - 1])) break;
+            rb_ary_push(path, values->ptr[child_head - 1]);
         } else if (frame->phase == JSON_PHASE_OBJECT_COMMA && count >= 2) {
-            if (!json_path_append_key(path, values->ptr[child_head - 2])) break;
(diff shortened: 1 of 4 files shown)


byroot:
Code looks great, I'll merge it soon.

> Happy to make it point at the object instead if you'd rather keep them consistent.

No I think this was the good call. I might even change it for the `line/column` too, as I think it can simplify the parser a little bit.

>  `on_load`-transformed keys are handled.

Hum, I need to check this a bit deeper, because in theory `on_load` can return about anything, so in some cases you just can't handle it. I'd be fine with just not generating the path in such case, but I'll dig a bit deeper.

> It also works for ResumableParser, where line/column can't be accurate:

Yep, I like that.

> Can look at that as a follow-up if there's interest.

Yes, adding position and json_path to the Java parser would be welcome in a followup.

byroot:
> , because in theory `on_load` can return about anything, so in some cases you just can't handle it.

I forced push your branch with an extra `on_load` test that show how it can be broken:

```
Failure: test_parse_error_json_path_on_load(JSONParserTest): [JSON::ParserError] exception expected, not #<NoMethodError: undefined method 'inspect' for an instance of BasicObject>.
test/json/json_parser_test.rb:949:in 'JSONParserTest#assert_parse_error_at'
test/json/json_parser_test.rb:864:in 'JSONParserTest#test_parse_error_json_path_on_load'
     861:   end
     862: 
     863:   def test_parse_error_json_path_on_load
  => 864:     assert_parse_error_at "$.a.b.c" do
     865:       JSON.load('{"a": {"b": {"c":', -> (obj) {
     866:         if String === obj
     867:           BasicObject.new
```

I think we should just not build the path it if a Hash key isn't a String/Symbol or an array index isn't an Integer.
Alternatively we could just not build it at all when `on_load` is passed, but that might be a bit too extreme.

ojuschugh1:
Went with your first suggestion: the path stops at the first key that isn't a String or Symbol, so the valid prefix is kept (`$.a` if only `"b"` was replaced by the proc, `$` if all keys were). That also removes the inspect fallback, so building the path can't call into user code at all anymore. Extended your test with the partial case.

ext/json/ext/parser/parser.c
```diff
@@ -651,11 +651,72 @@ static VALUE build_parse_error_message(const char *format, JSON_ParserState *sta
...
+        char c = RSTRING_PTR(key)[i];
+        bool alpha = (c >= 'a' && c <= 'z') || (c >= 'A' && c <= 'Z') || c == '_';
+        plain = alpha || (i > 0 && c >= '0' && c <= '9');
+    }
+
+    if (plain) {
+        rb_str_cat_cstr(path, ".");
+        rb_str_cat(path, RSTRING_PTR(key), len);
+        return true;
+    }
```
byroot: I just had the realization that it would probably be simpler to just push the raw objects in an array, and build the `json_path` string in Ruby. That would also help share the code with the eventual JRuby version.

ext/json/ext/parser/parser.c
```diff
@@ -651,11 +651,42 @@ static VALUE build_parse_error_message(const char *format, JSON_ParserState *sta
...
+
+    for (long depth = 1; depth < frames->head; depth++) {
+        json_frame *frame = &frames->ptr[depth];
+
+        bool innermost = depth == frames->head - 1;
+        long child_head = innermost ? values->head : frames->ptr[depth + 1].value_stack_head;
+        long count = child_head - frame->value_stack_head;
+
+        if (frame->type == JSON_FRAME_ARRAY) {
+            rb_ary_push(path, LONG2NUM(frame->phase == JSON_PHASE_ARRAY_COMMA ? count - 1 : count));
```
byroot, line 669: TODO: not even sure we need that decrement here.

byroot: Nevermind, I get it.

PR #1068, opened 2026-09-03
Fix SEGV when as_json returns a non-String for an invalid symbol

Watson1978:
In strict mode, a Symbol whose name is not UTF-8 compatible reached `raw_generate_json_string` together with whatever `as_json` returned, so a non-String result was read as a String. Depending on the heap layout that either segfaults or copies unrelated heap memory into the generated JSON, and the string path failed the same way once `as_json` returned another invalid symbol.

Strict symbols now go through the `T_STRING` case of `generate_json_general`, which already re-dispatches an `as_json` result by type and uses its `as_json_called` flag to stop a returned symbol from recursing.

## Reproduction

```ruby
require "json"

coder = JSON::Coder.new { |object, _| object.is_a?(String) ? object.bytes : object }
p coder.dump(["\xFF".b.to_sym])
```

Over 10 runs on json 2.21.2 with CRuby 4.0.6, this segfaulted 8 times:

```
common.rb:1089: [BUG] Segmentation fault at 0x0000000000000000
```

The other 2 runs printed heap memory instead of crashing, here with unrelated interpreter strings visible inside the JSON string:

```
"[\"\\u0007\xA8\\u0002...can't get const \\u0000Module#ruby2_keywords\\u0000...
...https://github.com/ruby/ruby2_keywords\\u0000...lib/ruby2_keywords.rb...\"]"
```

json 2.18.0, 2.19.3, 2.20.0, 2.21.1 and 2.21.2 all crash on the same input. With this change the script prints `"[[255]]"`, matching what the equivalent plain string already produced.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/ext/json/ext/generator/generator.c
+++ b/ext/json/ext/generator/generator.c
@@ -1124,15 +1124,6 @@ static void generate_json_fallback(FBuffer *buffer, struct generate_json_data *d
     }
 }
 
-static inline void generate_json_symbol(FBuffer *buffer, struct generate_json_data *data, VALUE obj)
-{
-    if (data->state->strict) {
-        generate_json_string(buffer, data, rb_sym2str(obj));
-    } else {
-        generate_json_fallback(buffer, data, obj);
-    }
-}
-
 static void generate_json_null(FBuffer *buffer, struct generate_json_data *data, VALUE obj)
 {
     fbuffer_append(buffer, "null", 4);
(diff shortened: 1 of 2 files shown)


ext/json/ext/generator/generator.c
```diff
@@ -1250,7 +1246,12 @@ static inline void generate_json_general(FBuffer *buffer, struct generate_json_d
                 }
                 break;
             case T_SYMBOL:
-                generate_json_symbol(buffer, data, obj);
+                if (!data->state->strict) {
+                    generate_json_fallback(buffer, data, obj);
+                } else {
+                    obj = rb_sym2str(obj);
+                    goto start;
```
byroot: I think we could use a fallthrough into T_STRING.

PR #1073, opened 2026-09-08
Update documentation to match json 3.0 API

bquorning:
json 3.0 changed `JSON.parse` to take keyword arguments instead of a positional opts Hash, so some of the documented usage would now raise `ArgumentError`. I have updated all such examples to use kwargs instead.

Also align documented output with what is actually produced with the current parser/generator on Ruby 4.0, plus a number of other minor changes.


byroot (approved):
❤️

